Issue #12, opened 2026-08-17
Guided vs unguided decomposition on MuSiQue — informs router design (router stays per ADR 0010)


AhmadiJahid:
**Outcome.** Run the decomposer on the MuSiQue eval set both ways — guided (hop count supplied) and unguided (model decides) — under the fixed few-shot method of ADR 0006, and log both to `experiments/log.md`.

**Why it matters.** ADR [0006](docs/adr/0006-drop-the-jury-fix-dataset-roles-and-the-few-shot-method.md) item 6 leaves "whether the pipeline keeps the router" OPEN and names this experiment as what decides it. Everything about the router — including the cap conflict in #10 — waits on it.

**Done when.** Both arms are evaluated on the same eval set (ADR 0007, 600 questions / 200 per hop) with a log entry and metrics JSON for each, and the comparison is on the table for Jahid.

AhmadiJahid:
Lane state (durable handoff, 2026-08-17 late morning): PR #21 implements the three-condition harness (unguided / oracle_guided / unguided_capped) — 122 condition checks, 29/29 smoke stages, dry run loads exactly 200/200/200 of the pinned 600 (ADR 0007). Independent code-review in progress; one open validity question under review: the shared max_new_tokens=256 may truncate the unguided arm's runaway-length behavior (spec puts the token cap only on unguided_capped). After review+fix: merge, then build the MuSiQue retrieval artifact over the pinned 600 (pool 2000, typed, bi-encoder top-20 → CE top-5) from committed configs, then run all three arms (log entry first, fixed seed, tmux; outputs under /tmp while #18 blocks home-disk writes) and evaluate with --compare pairwise, per-hop. No metric claims yet — nothing has run.

AhmadiJahid:
Reframed per ADR 0010 (2026-08-18 supervisor decisions, relayed by Jahid): the router **stays** in the pipeline, so this experiment no longer decides keep/drop. It still runs unchanged — same three conditions, same hypothesis — but its role is now to inform the router's design: how much a correct hop count actually helps the decomposer, and whether runaway generation length is the dominant failure. Implementation is in flight on `feature/12-guided-vs-unguided`.

AhmadiJahid:
Measured note for whoever runs this (retrieval input, not the harness).

[...]

PR #21, opened 2026-08-17
feat: guided vs unguided decomposition conditions on the MuSiQue eval set

AhmadiJahid:
Refs #12 — the harness only. The issue closes when the experiment has run and reported.

## What this adds

Three run conditions over the pinned v1 600-question MuSiQue set (ADR 0007:
`musique/dev_data/musique_ans_v1.0_dev_sample_{2,3,4}_hop_200.jsonl`), sharing model, seed,
decoding and retrieval so the only difference between them is the hop information in the
prompt and the step-line budget — **enforced by the runner, not just claimed here.**

| condition | prompt | generation |
|---|---|---|
| `unguided` | no hop count | `generation_overrides.max_new_tokens` only |
| `oracle_guided` | the query's gold hop count + each exemplar's own gold hop count | same |
| `unguided_capped` | no hop count | stops after `stop_after_step_lines` step lines (default 8) |

Selected with `--condition <name>`; `configs/decomposer_musique.json` names `unguided` as the
default. A condition block may set only `guided` and `stop_after_step_lines` — any other key is
refused, so no arm can move the model, the seed or decoding, and `--guided` may not contradict a
named condition. `configs/decomposer.json` keeps the MetaQA defaults.

Conventions recorded in **ADR 0013**, each sourced to Jahid's plan (prompt 4) or to the config.

## Round 1 fix pass (1 Critical, 7 Major, 6 Minor) — re-reviewed and passed

**Critical — the arms differed in more than hop information.** `prompt_unguided.md` is now
**derived**: the guided prompt minus its hop-bearing lines, byte for byte, nothing added. v1's
files differed in more than the hop count (mistral swapped the hop rule for "Decompose into the
minimal number of atomic steps."; qwen dropped the entire 7-rule block), so both were
regenerated and are **no longer byte-identical to v1** — a deliberate, recorded deviation from
the porting convention (ADR 0013, config `_unguided_prompt_note`, component README). A guard
makes any residual line-level delta a refusal with a diff.

**Major** — (1) `retrieval.require_input`: no silent fallback to random MetaQA exemplars (ADR
0006). (2) the pinned evaluation set is asserted on the rows actually loaded, with an explicit
`--allow-unpinned-eval-set` opt-out for fixtures, and an unparseable hop id is a refusal rather
than `hop_fallback`. (3) one step normalization (`src/step_lines.py::split_step_lines`) shared
with the evaluator, counted after `<think>`/tail post-processing. (4) `rows_at_max_new_tokens`
per arm. (5) honest reporting: only one model folder can run these arms. (6) self-exclusion on
every few-shot path. (7) `prompt_sha256` + `retrieval.input_sha256` in the snapshot.

**Minor** — cap described as `StoppingCriteria` + trim; the hop guard is a line regex; identical
per-row dump fields in all arms; the criterion decodes only the new tail.

## Round 2: main merge + delta re-review findings

**Merged `main` (PR #24, LoRA arm) — nine conflicts.** The load-bearing one is `generate()`,
which both PRs rewrote; the resolution keeps **both** #24's cost accounting (prompt/completion
tokens, CUDA-synchronized latency around `model.generate` only, `cost_summary`) and #12's
`post_cfg` + `max_step_lines` cap. One call per row, one dict out, cost and truncation fields
side by side. The two branches had also added the same path template under two key names —
converged on main's `musique_eval_questions_template`. The add/add fixture eval files keep 3
rows per hop **and** main's 1-per-hop ids, so the LoRA train/eval overlap assertion still has
its rows (`smoke_paths.eval_set_expected` → 3/9, and the two LoRA expectations that read 1/3
updated). This branch's ADR is renumbered **0012 → 0013** because main took 0012.

**New findings fixed**

- **I1 — the pinned set was counted, not identified.** The loaded question ids are now compared
  against the ids of the three ADR 0007 files (`datasets.musique_eval_questions_template`),
  refusing on mismatch with the offending ids named (capped at ten). Counts alone let a decoy
  600 through. The config prose and the ADR bullet now describe identity, and the metrics carry
  `evaluation_set.id_identity_checked` / `ids_missing_count` / `ids_unexpected_count`.
- **I3 — over-claimed causality.** `rows_stopped_at_step_line_cap` is read from whether the
  `StoppingCriteria` actually fired (`StepLineCapState`), not from the step count, so a model
  that ends on exactly the cap by itself is no longer reported as capped.
  `rows_at_step_line_cap` keeps the output-shape meaning; both are defined in `metrics.json`'s
  `truncation_definitions`.
- **N1 — whole-line derivation.** A guided line mixing a hop reference with another instruction
  is refused ("split the line"): otherwise the derivation would drop that instruction from the
  unguided arm and the byte-equality guard could not see it, since it compares against the same
  faulty derivation.
- **N2** — snapshot `few_shot_self_exclusion.enabled` reflects `few_shot_enabled`.
- **N4** — the rows-loaded comment now states the truth (after `sample_size`, before the
  `--dry-run` limit).

**I2 — resolved by Jahid on 2026-08-19 (issue #12): option (a).** In a guided prompt each
few-shot exemplar's `Hop count:` line now states **that exemplar's own gold hop count** — the
number of steps in its own gold decomposition (`pool_few_shot_decomposition_musique`), counted
with `src/step_lines.py::split_step_lines`. The query's hop line keeps the query's gold hop
count; the unguided arms stay hop-free everywhere. An exemplar with a missing or empty gold
decomposition is **refused**, naming its pool id — no fallback to the query's hop count, since
that fallback is exactly what the decision removes.

Measured effect on the three 600-row dry runs: exemplar hop lines contradicting their own
decomposition go from **1740 / 3000 (58.0%)** to **0 / 3000**. v1 stamped the query's hop count
on every exemplar, so **v1's guided MuSiQue slide numbers are not directly comparable** to v2's
guided arm; recorded in ADR 0013 (item 8), together with the two options not taken (drop the
exemplar hop lines in both arms; keep v1's behaviour and state the confound).

**The MetaQA path is untouched.** `format_few_shot_examples` is shared, so the rule is selected
per config by `few_shot_exemplar_hop_count` (required in both configs, no default in code):
`configs/decomposer_musique.json` sets `exemplar_gold`, `configs/decomposer.json` keeps v1's
`query`. Verified: a guided MetaQA dry run renders **byte-identical** prompts before and after
the change (2 prompts, 6 hop lines each, all the query's hop), and a test now locks that in.

## Round 3: the config now runs the pinned experiment (ADR 0014)

Jahid decided on 2026-08-19 (**ADR 0014**, on `main`) that issue #12 runs on the **v1 pool** via
the existing v1 retrieval artifact, unchanged: `sim_dev_sample600_top20_rerankTop5.jsonl`, 600
rows, sha256 `e5c418a9b25f…`. That artifact is now named **in the config**, so an arm needs
nothing but `--condition` — no `--retrieval-input` to remember, and no way for two arms to read
different exemplars because someone typed a different path.

- `configs/decomposer_musique.json` → `retrieval.input_key: "musique_few_shot_top5_pinned600"`,
  which resolves through `datasets.*` in `configs/paths.json` against `data_root` — the repo
  convention for data outside the tree; **no absolute path in a config**. `require_input: true`,
  `hop_fallback: null` and the sha256 recording are unchanged.
- Resolution order in `resolve_retrieval_input`: `--retrieval-input` → `retrieval.input` →
  `retrieval.input_key`, resolved lazily so the fixture paths config is unaffected. Setting both
  config fields is a **refusal** (one config, one retrieval file); setting none under
  `require_input` stays the ADR 0006 refusal. The snapshot now also records `retrieval.input_key`
  beside the resolved path and its hash.
- **The pool is 9,156, not 2,000.** ADR 0014 records that deviation from ADR 0006 §4 as Jahid's
  call pending supervisor confirmation, so the stale "pool 2000" claims are gone from the
  refusal message and from ADR 0013 item 7, which now points at ADR 0014 for both the artifact
  and the pool. Thesis reporting states 9,156 wherever these results appear.
- `main` merged into the branch (ADR 0014 + the `exp-001` log entry; one index-line conflict).

## Evidence

Run with `/cta/users/fyilmaz/Thesis---QAv2/.venv/bin/python`, `PYTHONPYCACHEPREFIX` and all run
output under `/tmp`. A LoRA training run was live on the GPU throughout; every check here is a
dry run that loads no weights.

- `python -m unittest discover -s tests` — **84 tests, OK** (includes main's
  `test_finetune_lora.py`, i.e. the LoRA side still passes after the `generate()` resolution).
- `python scripts/smoke_test.py` — **33/33 stages passed** (both branches' stage groups).
- `python tests/test_decomposer_conditions.py` — **261 checks passed** (249 with
  `--skip-data-checks`); 253 / 243 before Round 3, which added six direct checks on
  `resolve_retrieval_input` (resolution order, the both-fields-set refusal, the `require_input`
  refusal — no longer reachable through the committed config — and `require_input: false`
  returning `None`) and two data-gated checks that the pinned artifact exists with one row per
  evaluation question. Covers: the compound-hop-line refusal; the cap-state flag (fired /
  not-fired); the per-row cost+truncation field set; the **decoy refusal** — same per-hop
  counts, relabelled ids, refused, and permitted only under `--allow-unpinned-eval-set`, then
  recorded as `pinned: false` with `ids_unexpected_count: 9`; both exemplar-hop modes with
  hand-computed counts plus the three refusal shapes; the rendered oracle_guided prompts'
  exemplar hop lines against their own decompositions; and the MetaQA guided prompt staying on
  v1's rule.
- **Three dry runs against the real read-only `data_root`**, one per condition,
  `--dry-run-limit 600`, same `--retrieval-input` (v1
  `sample_extracts/sim_dev_sample600_top20_rerankTop5.jsonl`, sha256 `e5c418a9b25f…` — the same
  artifact ADR 0014 has since pinned in the config, so these runs read what the real arms will):
  - `rows_loaded_total 600`, `{"2": 200, "3": 200, "4": 200}`, 600 distinct ids,
    `pinned true`, **`id_identity_checked true` with `ids_missing_count 0` and
    `ids_unexpected_count 0`** against `pinned_id_count 600`;
  - hop mentions in the 600 logged prompts: **0/600** `unguided`, **0/600** `unguided_capped`,
    **600/600** `oracle_guided`;
  - `unguided` and `unguided_capped` prompts byte-identical; `oracle_guided` minus its
    hop-bearing lines byte-identical to `unguided` for all 600;
  - **exemplar hop lines: 0 / 3000 contradict their own decomposition** (1740 / 3000 before
    Jahid's decision); the query hop line equals the row's gold hop count **600 / 600**;
    the stated exemplar hop values spread 2:1244 / 3:1424 / 4:332, i.e. they follow the
    exemplars rather than the queries;
  - identical id order across arms, identical `retrieval.input_sha256`, cap 8 / None / None,
    every arm recording `few_shot_exemplar_hop_count: exemplar_gold`.
  - **Negative control at the same scale:** the 600-row decoy (identical rows, ids relabelled
    `<hop>hop__decoy_N`) exits **1**, naming 600 missing pinned ids and 600 unexpected ids,
    while its per-hop counts are 200/200/200 — the counts-only check would have passed it.
- **Round 3 — one dry run with NO `--retrieval-input`**, i.e. the config default resolving on its
  own: `--model mistral_7b_instruct --config decomposer_musique.json --condition unguided
  --dry-run --dry-run-limit 600`. Observed: `rows_loaded_total 600`,
  `{"2": 200, "3": 200, "4": 200}`, 600 distinct ids, `pinned true`,
  `id_identity_checked true`, `ids_missing_count 0`, `ids_unexpected_count 0` against
  `pinned_id_count 600`; snapshot `retrieval.input_key: musique_few_shot_top5_pinned600`,
  `input_resolved` = `<data_root>/musique/sample_extracts/sim_dev_sample600_top20_rerankTop5.jsonl`,
  `input_sha256: e5c418a9b25f5ef290a78dd3a99642e4b0c5d7b6fbe6b5e2785a1424239ae70d`; **0 of the
  600 logged prompts mention a hop count** (the 9 files containing the substring "hop" are
  question text — "Chopin", "Hope", "Hopewell", "hoping"). Dry run: no weights loaded, so
  quality and the parameter-count assertion are still unexercised.
- MetaQA path unchanged; `--condition` against `configs/decomposer.json` still refused.

Parameter-count assertion is **not** exercised (all dry runs). Decomposition quality is
**unmeasured**; no experiment-log entry is claimed.

## Open questions for Jahid (ADR 0013)

1. The comparison rests on **one model** (`mistral_7b_instruct`) — `qwen3_5_9b` is 9B, above the
   ceiling. Adding an unguided prompt to another ≤8B folder is your decision.
2. Is `generation_overrides.max_new_tokens: 256` the intended token cap? Unmeasured plumbing
   default, and it bounds the uncapped arms.
*(The third — what the exemplar hop lines should say — you decided on 2026-08-19; it is
implemented here and recorded as ADR 0013 item 8.)*


AhmadiJahid:
Gate-1 review (independent code-reviewer, isolated worktree at 768cfb9): **Critical-block** — 1 Critical, 6 Important, 5 nits. Full detail in the session record; the load-bearing items:

**C1 (Critical, blocks):** `oracle_guided` stamps every few-shot exemplar with the *query's* hop count (`format_few_shot_examples`, run_decomposer.py:284-297), so a 2-hop query shows exemplars labeled 'Hop count: 2' whose decompositions have 1–3 steps — every exemplar contradicts the prompt's 'steps MUST equal hop count' rule. The oracle arm is systematically weakened, biasing the router-informing experiment toward 'guidance doesn't help'. Pre-existing v1-parity code, so the resolution is Jahid's (Gate 3): (a) clean oracle — stamp each exemplar with its own step count or drop the exemplar hop line, diverging from v1; or (b) keep v1 parity and record the confound in config, log hypothesis, and an ADR.

**Important:** I2 — a MuSiQue run without `--retrieval-input` silently falls back to random MetaQA exemplars (launch precondition + should be a refusal); I3 — the cap's line rule and the evaluator's `_split_decomposition_text` are unpinned duplicates that demonstrably diverge on 6 separator classes (fix: one shared split function + byte-identical test; alt branch c1b4aed has a working reference); I4 — nothing records token-budget exhaustion (add per-row stop reason: EOS / budget / step cap), and uniform max_new_tokens=256 vs the plan's capped-arm-only reading is Jahid's to settle; I5 — unguided prompt carries a 'minimal atomic steps' rule line the guided prompt lacks, so the contrast is not hop-info-only (align or record); I6 — qwen3_5_9b is 9B, refused by the 8B ceiling, so exactly one model folder (mistral_7b_instruct) can run these arms; I7 — `hop_fallback: 2` would fabricate an oracle on an unparseable id (should hard-error in guided arms).

**Cleared:** post-hoc truncation exists and binds on the runner's rule (the alt-lane concern was stale); max_new_tokens applies identically to all arms — the unguided arm is NOT differentially throttled; MetaQA path intact (300 questions, rc=0); seeds/ceiling/eval-set/no-data-in-git all pass; no PR #22 incompatibility (textual smoke_test.py offset at most).

Reviewer evidence: 122 condition checks + 8 metric tests + 29/29 smoke stages at the PR SHA; dry runs of all three arms against the real data_root (600 rows, 200/hop, seed 42, commit-stamped).

Issue #1, opened 2026-08-11
Confirm or replace the idea paragraph in README.md


turalnovruzov:
The description of the research in `README.md` was transcribed by Codery from meeting audio, not written by Jahid. Meeting transcription is unreliable and this is his thesis, so the paragraph needs his word before anything downstream leans on it.

**Outcome:** the README paragraph says what Jahid says it says — either confirmed as-is or replaced with his own wording.

After that it is fixed: agents never extend or edit it, and a domain fact that is not in it gets asked about rather than derived (`CLAUDE.md`, There is no PRD).

PR #22, opened 2026-08-18
feat: harden the decomposition comparison tooling and pin step normalization (#20, #9)

### LINKED ISSUE #9: Smoke test: add a punctuation-only-variance fixture row so normalization drift turns the test red
Follow-up from PR #7 review (finding N1, non-blocking). The 13 golden metric assertions catch most metric-semantics drift (verified: mutating the step-F1 TP rule and the step-enumerator strip both turn the smoke test red), but deleting the punctuation-strip line in `_normalize_step` (`scripts/musique_decompositions_evaluator.py:121`) leaves all golden values green — the current fixtures contain no punctuation variance between predicted and gold steps.

Fix: add one fixture prediction whose step differs from gold *only* by punctuation, so the documented normalization rule (lowercase, punctuation stripped except `#`, whitespace collapsed) is itself pinned by a golden value.

Small, self-contained; suits the next ml-engineer pass that touches tests.

### LINKED ISSUE #20: Harden the comparison tooling: six minor findings from PR #17's review
PR #17 passed review (stats verified against independent reference implementations; merged as c9d525c). Six non-blocking findings for a later pass, with the reviewer's fix directions:

1. **Two gold denominators**: directional metrics use len(gold_steps); hop-count metrics use the gold hop_count field. Real gold has 0 disagreements today (2417 rows scanned) — assert equality at gold load, or document both denominators in docs/METRICS.md.
2. **No step_count_exact_rate**: 1 − over − under is not emitted; readers will wrongly reach for hop_count_exact_match_rate (different denominator per 1). Emit it or document its absence.
3. **Directional metrics don't reach all_runs.csv**: pool_sweep_orchestrator SUMMARY_FIELDS not extended — but the sweep CSV is exactly where the supervisor's under-decomposition question will be asked. Extend SUMMARY_FIELDS.
4. **--compare recomputes composite with current config weights**, undetectably different from the weights the per-item files were scored under. Stamp weights into the per-item file and refuse on mismatch (like the id-mismatch refusal).
5. **Bootstrap memory is O(iterations × n) unchunked** — fine at n=600, OOMs around n≈10k. Chunk or document the ceiling.
6. **`significant` emitted with no floor on n** (n=3 still prints a boolean; McNemar with b+c=3 cannot reach 0.05). Record n and minimum attainable p, or warn below a config threshold.

Nits: run-note table header says CI where McNemar rows carry p-values; METRICS.md "single exception" sentence has two more exceptions.

ADR 0009 records the protocol itself (pending supervisor confirmation). Done when the six are fixed or explicitly declined here.

AhmadiJahid:
Six non-blocking findings from PR #17's review, applied in the reviewer's fix direction, plus the punctuation fixture from issue #9. Everything extends the existing evaluator / orchestrator — no parallel tooling, no new script.

## Issue #20, finding by finding

1. **Two gold denominators — assert *and* document.** `_load_gold` now aborts when a gold row's positive `hop_count` disagrees with `len(question_decomposition)`, listing the offending ids (capped by the new `gold_validation.max_reported_mismatches`); rows without a usable field still fall back to the step count, where the two agree by construction. Both denominators are documented in `docs/METRICS.md` §1. The 2417-row scan reported in issue #20 is cited as PR #17's measurement and **not** re-run here.
2. **`step_count_exact_rate` emitted.** In the aggregate and in every `per_gold_hop_metrics` block (one `_aggregate`, so per-hop comes for free), counted directly as `#(signed == 0) / m`, which is identically `1 − over − under`. `METRICS.md` states how it differs from `hop_count_exact_match_rate` (different denominator) and `METRIC_DEFINITIONS` carries the same warning into every metrics JSON. Also added to the run note's step-count line.
3. **Directional metrics reach `all_runs.csv`.** `SUMMARY_FIELDS` gains `step_count_mae`, `mean_signed_step_count_error`, `over_decomposition_rate`, `under_decomposition_rate`, `step_count_exact_rate` (`_METRIC_FIELDS` is derived by exclusion, so they populate automatically). One addition beyond the finding, because the finding created the hazard: appending new columns to a table written with the old header would silently misalign every value, so `_append_summary` now refuses on a header that does not match `SUMMARY_FIELDS`, in the same style as the existing dev-identity refusal.
4. **Composite weights stamped and checked.** `<prefix>_per_item.json` is now an object — `schema`, `created_utc`, `predictions_path`, `gold_path`, `composite_score_weights`, `composite_step_count_error_scale`, `items` — and `--compare` refuses when the two files disagree on weights or scale, printing both (same loud style as the id-mismatch refusal). A legacy bare-list file is refused with an instruction to re-run. The config's weights (which the bootstrap composite is still recomputed with, unchanged behaviour) are no longer *undetectably* different: the metrics JSON records `per_item_composite_score_weights`, `per_item_composite_step_count_error_scale` and `config_weights_match_per_item_files`, and the run note prints a WARNING line when they disagree. **Format break:** old per-item dumps under `runs/` must be regenerated; nothing else in the repo reads that file.
5. **Bootstrap memory bounded.** Resample indices are drawn in `paired_comparison.bootstrap_chunk_size` (1000) blocks and reduced immediately, so peak memory is O(chunk × n) instead of O(iterations × n); only the per-statistic difference vectors survive. **Chunking changes no number**: `rng.integers` consumes one draw per index in row-major order, so chunked draws concatenate to the single-block draw. Verified two ways below; no golden value moved, so nothing had to be re-pinned.
6. **Significance floor recorded.** Every bootstrap and McNemar row now carries `n` and `underpowered`; McNemar rows also carry `min_attainable_p_value` (`min(1, 2·0.5^m)` for m = b + c discordant pairs, 1.0 at m = 0) and `min_attainable_p_reaches_alpha` — with m = 3 the floor is 0.25, and m = 6 is the smallest count that can reach p < 0.05 at all. A top-level `significance_floor` block records n, the threshold, `below_min_items` and a `warning` string, which the run note prints.
   **Decision for Jahid:** `paired_comparison.min_items_for_significance_claim` is set to **30**. That number is mine, not the supervisor's; the config `_note` and `METRICS.md` both label it a reporting guard rather than a statistical standard, and it is a one-line config change if he or his supervisor wants a different floor (or none).

**Nits.** The run-note interval column is headed `CI or p` (bootstrap rows carry an interval, McNemar rows a p-value), with a new `underpowered` column. `METRICS.md`'s "single exception" sentence now names all three non-macro keys: `reference_validity_micro`, the two hop distributions, and `composite_score`.

## Issue #9 — punctuation normalization pinned

New fixture gold row `2hop__d004_p` (2 hops, `hop_count` consistent with its 2 steps) and a prediction whose steps differ from it **only** by punctuation (`permit?` → `permit.`, `[#1]?` → `[#1]`). Hand-computed: both steps normalize identically, so exact match / step P,R,F1 / ordered accuracy are 1.0 and the composite is exactly 1.0, while `_tokenize` (which does not strip punctuation) sees 8 of 10 LCS tokens, so ROUGE-L P = R = F1 = 0.8. Fixture aggregates were recomputed by hand and re-pinned: `total_evaluated` 4, `exact_match_rate` 3/4, `step_f1_macro` 11/12, `rouge_l_f1_macro` 51/55, `composite_score` 113/120, plus `per_gold_hop_metrics.2` (`exact_match_rate` 1.0, `step_f1_macro` 1.0, `rouge_l_f1_macro` 0.9) and the compare point estimates (11/12, 113/120).

## Evidence (ran-X-observed-Y)

- `.venv/bin/python -m unittest tests.test_decomposition_metrics tests.test_pool_sweep_summary` → **23 tests, OK** (19 in the metrics file, 4 new in the sweep-summary file).
- `.venv/bin/python scripts/smoke_test.py` → **25/25 stages passed** at this commit, run in a clean `git worktree` checkout of it so the result depends on nothing uncommitted.
- **Issue #9 red/green:** deleted line 226 (`x = _PUNCT_KEEP_HASH_RX.sub("", x)`) → `smoke_test.py --only decomposition_metric_tests,musique_eval` **FAILED 0/2**, with `exact_match_rate` 0.75 → 0.5, `step_f1_macro` 0.9167 → 0.6667, `composite_score` 0.9417 → 0.7667, `per_gold_hop_metrics.2.exact_match_rate` 1.0 → 0.5, and 7 unit-test failures including the new `TestStepNormalization` (composite 0.3 vs 1.0). Restored the line → **PASS 2/2**.
- **Chunking equivalence:** transcribed the pre-change code path (one `rng.integers((10000, n))` draw) and compared its CIs against `_paired_bootstrap` at chunk sizes 1000, 1, 7 and 10000 on the fixture per-item file (n = 4, seed 42, a degraded system B): **bit-identical in all four cases** (`ci_low`/`ci_high` equality, not approximate). `TestBootstrapChunking` pins the same invariance in CI, including against `chunk_size == iterations` (the un-chunked path) on a non-degenerate setup, plus the raw index-stream concatenation and the refusal of a non-positive chunk size.
- **Compare path exercised end to end** on the fixture with a real difference: table prints the new `CI or p` / `underpowered` columns, `min attainable p=0.5` for b = 2, c = 0 (hand-check: `2·C(2,0)/2²` = 0.5), and the n = 4 < 30 warning. A weights-differ-from-config run printed the expected WARNING and `config_weights_match_per_item_files: false`.
- New hand-computed tests: `TestStepNormalization`, `TestGoldDenominators` (tampered gold refused, naming `2hop__d001_a` and `hop_count=3, steps=2`), `TestMcNemarPower` (b=3, c=1 → p = 0.625, min p = 0.125), `TestBootstrapChunking`, `test_n_below_the_floor_is_flagged`, `test_weight_mismatch_between_files_is_refused`, `test_legacy_bare_list_per_item_file_is_refused`, and `step_count_exact_rate` added to every existing directional assertion block.

No GPU work, no model loaded, no data added to git (both fixture edits are fabricated rows). No experiment was run, so `experiments/log.md` is untouched.

**Note for the reviewer:** this branch was cut from `8038f5b` in a working tree that also holds another lane's uncommitted issue #12 work (step cap / MuSiQue conditions). Only the files above were staged — `scripts/smoke_test.py` was staged hunk-by-hunk so the other lane's three new smoke stages are not in this commit — and the commit was verified in isolation in a separate worktree.

Closes #20
Closes #9


AhmadiJahid:
Gate-1 review (independent code-reviewer, isolated worktree at b3c2bae): **approve — no Critical**. Verified: 23/23 tests, 25/25 smoke stages; bootstrap chunking independently confirmed bit-identical at real scale (10k×600) and at awkward chunk sizes; all fixture goldens re-derived by hand in exact fractions and matching; #9's red direction re-derived analytically (permission sandbox blocked the temporary line-delete) and confirmed pinned; McNemar min-attainable-p formula hand-checked; ADR 0009 substance unchanged; no closed-model surface, no data in git.

Two Important findings, neither blocking, both actioned outside this PR:
1. Three conventions established here need the ADR trail — the n≥30 significance-claim floor (implementer's provisional number), the versioned per-item schema `musique_decomposition_per_item/1` + format break, and the stamped-weights mismatch rule. → ADR 0011 records them as provisional pending Jahid/supervisor.
2. Finding 4 departed from the stated fix direction: file-vs-file weight mismatch **refuses**, config-vs-file only **warns** (recorded, exit 0). Deliberate, honestly labeled; whether to harden to refusal is Jahid's call — recorded in ADR 0011 as the provisional house rule.

Nits (non-blocking, for a future pass): header refusal fires mid-sweep after compute is spent (belongs up front at pool_sweep_orchestrator.py:676); the refusal message's `--overwrite` remedy doesn't actually work (appends duplicates under the stale header); both table guards degrade to allow on unreadable files; `schema` read but never validated; over+under+exact=1 identity silently fails on empty blocks (reads as 0% exact); `_load_gold` now whole-file + hard-abort with no escape for a future gold refresh; `step_count_mae`/`step_count_abs_error_mae` duplicate columns; committed example all_runs.csv fixture is two generations stale.

Issue #16, opened 2026-08-17
End-to-end evaluation: MuSiQue answering backend + MetaQA compile-execute wrapper


AhmadiJahid:
**Outcome.** Decompositions get scored on whether they lead to the right answer, not only on how they look: an answering backend for MuSiQue, and a wrapper that compiles sub-questions and executes them for MetaQA.

**Why it matters.** ADR [0006](docs/adr/0006-drop-the-jury-fix-dataset-roles-and-the-few-shot-method.md) gives MuSiQue both decomposition-quality and end-to-end evaluation, and MetaQA end-to-end only (no gold decompositions) through the supervisor's GRAG system. Without this, half the evaluation the ADR specifies does not exist.

**External dependency to chase:** GRAG is the supervisor's system — access, interface and how sub-questions are handed to it are unknown here and need Jahid to obtain them; that blocks the MetaQA half.

**Done when.** End-to-end answer accuracy for a decomposition run is measurable on MuSiQue, and the MetaQA path is either wired to GRAG or blocked with the dependency recorded.

AhmadiJahid:
**Follow-up carried out of PR #32 (review item I-2), deliberately not fixed there.** `components/decomposer/run_decomposer.py` reads `require(model_cfg, "chat_template.enable_thinking")` only inside the non-dry chat-template branch of its inference loop (line ~2424), so a chat-template model folder missing that key is a crash **no dry run can reach** — the exp-002/exp-003 failure shape of ADR 0016, in config form rather than result-key form. Both shipped chat folders carry the key today, so nothing is currently broken.

It was left alone in PR #32 because exp-004/exp-005 are running from that file's lineage and a fix there would have widened an answerer PR into the decomposer's launch path. The fix is one line: read it in a preflight before the model loads, the way `components/answerer/run_answerer.py::assert_generation_preflight` now does (`CHAT_TEMPLATE_KEYS`).

If this issue closes before the fix lands, promote this to its own issue rather than losing it.

PR #32, opened 2026-08-20
MuSiQue answering backend: score decompositions by the answer they lead to (Refs #16)

AhmadiJahid:
The **end-to-end** half of MuSiQue evaluation that ADR 0006 specifies and that did not exist. `Refs #16` — **not** `Closes`: the MetaQA/GRAG half of that issue stays open (it is blocked on the supervisor's system).

## What it does

`components/answerer/run_answerer.py` executes a decomposition and scores the answer it leads to:

1. reads the sub-questions in order — from a decomposer run's `results.json` (in that dump's actual schema), or from the gold decompositions for the **oracle-decomposition ceiling**;
2. substitutes each `[#k]` with the answer produced for step k (`src/step_lines.py::substitute_step_references`);
3. asks the reader each sub-question over the MuSiQue item's **full paragraph list**;
4. takes the **final** step's answer as the prediction;
5. scores it with MuSiQue's **official answer EM and answer F1** against the item's `answer` plus its `answer_aliases` (`src/answer_metrics.py`).

Metrics JSON: EM/F1 overall and **per gold hop depth (2/3/4)** in the reporting style of `docs/METRICS.md` §2, plus counts (items in source / missing item / evaluated / scored / no-steps / no-gold-answer, sub-questions prepared and executed, failed generations, references resolved and unresolved, truncations) and a cost block per sub-question. Aggregates and counts only — no dataset text in the metrics file; the per-item file, which does carry questions and answers, stays under the gitignored runs root.

## Methodology — decided by Jahid, not by this PR

ADR **0019** records the three decisions (Jahid, 2026-08-20, in session; **pending supervisor confirmation**): reader from the decomposer's registry (default `mistral_7b_instruct`, no new model family), full-paragraph context as the **only** policy that ships (`context.policy` refuses anything else), official EM/F1 against answer + aliases. Everything else is in `configs/answer_musique.json`. **No model scores, rates or judges anything** — EM/F1 are string metrics (CLAUDE.md standing constraint).

Three dataset facts in the ADR are measured, not assumed (2026-08-20): 2401 of 2417 dev rows carry 20 paragraphs and 16 carry 17–19; 680 of 2417 rows carry aliases, 136 of the pinned 600 (51/53/32 by hop); all 3,987 gold step references are bare `#k`, none bracketed — which is why the bare grammar is executed as well as `[#k]`.

## Guarantees

- parameter ceiling printed and asserted before any answer is generated (`src/model_size.py`, new `answerer` component key);
- the evaluation set is asserted to be the pinned ADR 0007 set **by id**, not only by count (the decomposer's own `assert_pinned_eval_set`, which gains an optional `remedy` sentence so a shared refusal does not send an answerer user to `--retrieval-input`); `--allow-unpinned-eval-set` is the recorded fixture opt-out;
- `--gold-decompositions` restricts itself to those pinned ids — a ceiling is a ceiling for one evaluation set (ADR 0011) — and reads only gold sub-question *text*, never gold sub-answers;
- every item's context is checked to exist **before** weights load: a scored item can never be silently closed-book;
- a reader prompt without `{context}` or `{question}` is refused;
- every config key the generation call reads is read up front, so a missing decoding or cleanup key cannot be a real-run-only crash.

## Verification (no GPU, no model loaded, no experiment launched)

The run lock is held by exp-004/exp-005 and the GPU is contended; nothing here loads weights.

- `.venv/bin/python -m unittest discover -s tests` → **195 tests OK** (measured **145** at `main` `ec1fad5`; +47 new checks in `tests/test_answer_musique.py` and +3 answerer negative controls in the generation-contract guard).
- `.venv/bin/python scripts/smoke_test.py` → **36/36 stages passed** (33 before the three new `answer_musique_*` stages), run at the PR SHA `bef65e0`.
- Dry runs observed end to end on the fixtures: predictions mode 5 items in → 1 unmatched → 4 evaluated, 11 sub-questions prepared, 6 `[#k]` references resolved, 0 unresolved, gold-hop distribution `{2: 2, 3: 1, 4: 1}`, EM/F1 `null` with an `unmeasured` note (a dry run generates nothing — not zero); oracle mode 9 pinned ids requested → 6 missing from the gold file → 3 executed, 9 sub-questions, one item per hop.
- **Per ADR 0016**, the real-run-only key contract is guarded statically: the answerer's dry run leaves `gen = None` for every step, so its `gen[...]` consumers and its cost copy are unreachable without weights. `tests/test_generation_contract.py::TestAnswererGenerationContract` ships negative controls (a renamed `answer` key, a phantom cost key), each asserting its replace-target matched.

## Reviewer notes

- The **reader prompt** (`components/answerer/prompts/reader.md`) is a mechanical extractive-QA instruction written here, not a recorded decision. Its wording is the one thing in this PR that a human may want to change.
- `src/answer_metrics.py` implements the SQuAD/MuSiQue definitions from their published form; this box has no network, so the upstream `metrics/answer.py` was **not** fetched. The provenance note says so, and the hand-computed vectors pin the behaviour — a reviewer with the upstream file should diff it before any number produced here is published.
- No experiment log entry: nothing was run. The first real run needs the run lock and its own `experiments/log.md` entry written before launch.


AhmadiJahid:
**Upstream diff verification of `src/answer_metrics.py` — done (lead session, 2026-08-20).** The provenance note's outstanding obligation is discharged:

Fetched from `github.com/StonyBrookNLP/musique` @ `main`: `metrics/answer.py` and `evaluate_v1.0.py`, archived (outside git, per data rules) at `/cta/users/fyilmaz/thesis-qav2-data/reference/musique_official/` — sha256 `10368f61…` (answer.py), `f5fe66ae…` (evaluate.py).

Compared function by function against `src/answer_metrics.py`:

- `normalize_answer`: identical semantics **and order** (lower → strip `string.punctuation` → article regex → whitespace collapse).
- `get_tokens`, `compute_exact`: identical.
- `compute_f1`: identical — multiset `Counter` intersection, empty-side rule (`int(gold_toks == pred_toks)`), zero-overlap → 0. Note the official `metric_max_over_ground_truths` passes `(prediction, ground_truth)` into functions whose parameters are named `(a_gold, a_pred)` — an upstream naming wart with **no numeric effect**, since both functions are symmetric in their arguments.
- Gold set: official `evaluate_v1.0.py:44-46` builds `[answer] + answer_aliases` — matches `gold_answer_set`, which additionally de-duplicates and drops empty/non-string entries (defensive; divergence only reachable on malformed gold data, where the official script would score an empty-string alias or crash on an empty `max()`).
- Aggregation: official `AnswerMetric` macro-averages EM/F1 over items — matches the documented aggregation.

**Verdict: numerically equivalent on well-formed data.** The provenance note in the module can be updated in the fix pass to cite this comment.

Also: **Jahid approved the reader prompt wording** (`components/answerer/prompts/reader.md`) as-is, 2026-08-20 in session — the one open human item the implementer flagged.

AhmadiJahid:
**Fix pass pushed: `fc8d572`.** All five `I-` items and N-1…N-4 applied on the same branch; not merged, delta re-review welcome.

| item | what changed | evidence |
|---|---|---|
| I-1 | `assert_reader_template` now splits on `chat_split_marker` and checks **the half that is actually filled** (`reader_filled_text`). The reviewer's repro (`{context}` only in the system half) is a negative control, next to a test that fills that template and asserts the user message contains no context. | `tests/test_answer_musique.py::TestContextAndPromptGuards` |
| I-2 | `assert_generation_preflight` reads `chat_template.enable_thinking` when `prompt_style == "chat_template"` (`CHAT_TEMPLATE_KEYS`); a test walks **every** folder in the registry. `run_decomposer.py`'s analogous gap untouched, filed on issue #16 (comment 5352822525). | `TestGenerationPreflight` |
| I-3 | New `answerer_max_params: 8e9` key + mapping; `load_limits` now requires every component key (`ceiling_for` only runs after weights load — the ADR 0016 shape). **Observed: the `qwen3_5_9b` smoke stage still passes, `ceiling_asserted: false` on the dry run**; a 9B reader is refused by test. No blocker. | `TestReaderCeiling` |
| I-4 | ADR 0019 gains an **"Implementation conventions (agent-authored)"** section: the no-abstention line (`prompts/reader.md:8`), the answer cleanup, the 64-token budget, the one-prompt-for-the-registry / `chat_split_marker` convention — and Jahid's 2026-08-20 approval of the wording. | `docs/adr/0019-musique-answering-backend-conventions.md` |
| I-5 | Provenance note cites **both** halves. I read the two archived files myself before writing it and confirm the wrapper: `normalize_answer` order, `get_tokens` / `compute_exact` / `compute_f1` (incl. `int(gold_toks == pred_toks)`), gold set `[answer] + answer_aliases` at `evaluate_v1.0.py:44-46`, `metric_max_over_ground_truths`, macro average in `AnswerMetric`. Consequences gain the `remedy` / `component` parameterization convention. | `src/answer_metrics.py` |
| N-1 | `[#1` is recognised, left verbatim, **always** counted unresolved (both grammars). | 3 tests |
| N-2 | The bare form requires standalone digits, so `#1st` is no longer a reference. Honest limit documented and tested: `Route #66` still parses as step 66 → left verbatim, counted, never rewritten. All 3,987 refs in the 2,417 gold dev rows are well-formed bare `#k` (measured), so real-data behaviour is unchanged. | 4 tests |
| N-3 | The shared pinned-set warning takes a `component` tag. | — |
| N-4 | A dry run truncated by `--dry-run-limit` records `pinned: false` with `pinned_on_full_load` and `rows_processed` beside it, so the preflight verdict survives without the run claiming the pinned set. Observed in the chat smoke stage: `rows_processed: 2 of 4`, reason string written. | metrics of `answer_musique_chat_template_dry_run` |

Two measured facts added while discharging I-5, both from `musique_ans_v1.0_dev.jsonl` (2417 rows) on 2026-08-20: **0** non-string aliases, **0** blank aliases, **0** rows with a duplicate inside `[answer] + answer_aliases`, **0** empty answers — so this repo's defensive de-duplication cannot diverge from upstream on this dataset; and **all 2417** rows are `answerable: true`, so upstream's answerable-only gating is not a difference either.

**Ran / observed at `fc8d572`:** `unittest discover -s tests` → **214 tests OK** (195 at the reviewed `bef65e0`, +19 for these items); `scripts/smoke_test.py` → **36/36 stages passed**. No weights loaded, no GPU touched, no experiment launched.

**One item I could not action:** N-5/N-6 were relayed as "docstring lines only, per the review" without their text — the review body is not on this PR, so I have no wording to apply. Relay the two lines and they are a one-commit follow-up.

*(This comment supersedes the previous one, which lost two table cells to shell backtick expansion.)*

Issue #30, opened 2026-08-20
Report a paired t-test alongside bootstrap CIs + McNemar


AhmadiJahid:
Jahid's call (2026-08-20, triaging #19): the supervisor's meeting ask was "a t-test, a statistical test"; #11 implemented paired bootstrap CIs + exact McNemar (ADR 0009). Rather than treating that as a silent substitution, **add a paired t-test to the reported battery** so the literal ask is covered — bootstrap + McNemar remain the headline protocol per ADR 0009.

**Done when.** The comparison artifact (`compare_decomposer_arms.py` / ADR 0011 conventions) reports a paired t-test next to the existing bootstrap CI + McNemar outputs, with the same pairing and the same eval set.

Small pipeline change (Gate 1 applies). Refs #11, #19, ADR 0009, ADR 0011.

Issue #6, opened 2026-08-12
Open research decisions carried from v1 — need Jahid + supervisor


AhmadiJahid:
Surfaced while carrying v1 knowledge into v2 (PR #5 review, 2026-08-12). These were implied by v1 but never explicitly decided/recorded, so no ADR exists. Each is a research-direction call — Jahid decides (with his supervisor where it touches methodology); when decided, it becomes one retroactive-free ADR + index row.

1. **Datasets:** MetaQA + MuSiQue as the thesis datasets (currently recorded only as *migration* scope in ADR 0001, not as a research decision).
2. **Router default:** Qwen2.5-3B few-shot + masked-similarity routing? Note: 1.5B/1.8B/3B candidates all exceed the ~600M router cap (standing constraint) — choosing any of them as-is needs the supervisor to change the cap, not an agent to ignore it.
3. **Decomposer prompting:** guided (hop-count-forced) vs unguided.
4. **Decomposer retrieval default:** typed masking + pool size ~2000–4000?
5. **Thesis-primary metric** (v1's composite is a weighted blend — see docs/prior-work.md for its definition and caveats).
6. **MuSiQue eval-set construction** (v1 used n=600, 200/hop from dev).
7. **Embedding model** for similarity retrieval.

No agent may resolve these; this issue is the durable home so they don't get lost in a closed PR description.

AhmadiJahid:
Status of these seven after the 2026-08-12 meeting and Jahid's 2026-08-17 call. Leaving this issue open — three items are still undecided.

**Decided by the 2026-08-12 Jahid + supervisor meeting — ADR [0006](docs/adr/0006-drop-the-jury-fix-dataset-roles-and-the-few-shot-method.md):**

- **Item 1 (datasets)** — decided. MuSiQue carries decomposition-quality *and* end-to-end evaluation (it ships gold decompositions); MetaQA carries end-to-end only, through the supervisor's GRAG system. (Same meeting also dropped the jury stage to Future Work, superseding ADR 0002.)
- **Item 3 (guided vs unguided)** — decided *how it gets decided*: by experiment, not by argument. Tracked in #12; its result also settles the router question.
- **Item 4 (retrieval default)** — decided. Typed entity masking, pool size fixed at **2000**, bi-encoder top-20 → cross-encoder re-rank to top-5. Raw masking, uniform masking and bi-encoder-only retrieval are now **ablations**, not live variables.

[...]

PR #33, opened 2026-08-20
Significance battery over v1's masking and retrieval comparisons (Refs #6)

AhmadiJahid:
Refs #6 (item 4 — decomposer retrieval default). Docs-only: one note at
`docs/analysis/2026-08-20-v1-masking-and-retrieval-significance.md`. **Do not merge yet** — for
Jahid's read before the supervisor conversation.

## Prior-work caveat (stated at the top of the note too)

Every input is a v1 artifact from `/cta/users/fyilmaz/Thesis---QA` (read-only, not modified). v1's
`runs/` is gitignored and untracked (`git ls-files runs/` returns 0), so these per-item files
**carry no commit SHA** — ADR 0005 territory. The results are **citable prior work, not v2
experimental claims**, and satisfy no part of Gate 2. Nothing was run on a GPU; there is no
`experiments/log.md` entry because no experiment was performed.

## Method

The house protocol (ADR 0009 as amended 2026-08-20: paired bootstrap + exact McNemar, with a paired
t-test alongside per #30), at v2's parameters — 10,000 resamples, chunk 1,000, alpha 0.05, seed 42,
one index matrix applied to both systems. v1's per-item files are bare lists without `item_id`, so
`--compare` refuses them; the battery was run from a session-local harness that **imports v2's own
`_paired_bootstrap` / `_statistic_arrays` / `_mcnemar_exact_p`** and was validated **bit-identical**
(`abs_tol=0`) against v2's implementation on the four house statistics, on one Task-A and one
Task-B pair. scipy 1.17.1 is present, so `ttest_rel` was used (no hand-rolled fallback).
Re-aggregating the v1 per-item rows reproduced each committed v1 `_metrics.json` to <= 3.3e-16.

## Headline findings

- **Masking, n = 600, item sets identical across all three variants (600/600, 200 per hop):**
  typed > raw is significant on exact match (+0.0217, McNemar p=0.0241), step F1 (+0.0232,
  CI [+0.0053, +0.0412]), ordered accuracy (+0.0213) and ROUGE-L F1 (+0.0126); hop-count EM is not.
  Bootstrap, t-test and McNemar agree 15/15 on the overall rows.
- **The composite gap that prior-work quotes (0.3606 vs 0.1888) is NOT significant**
  (CI [-0.1842, +0.2259]) and **87.3% of it is the single 0.2-weighted `reference_validity_micro`
  term**, which rests on 8 references from 3 items (typed) vs 4 from 1 item (raw) out of 600. With
  that term removed the typed-vs-raw advantage survives (+0.0273, CI [+0.0107, +0.0440]). At hop 3
  the term outright reverses the ranking.
- **typed vs uniform is undecidable at n = 600** — nothing significant on any metric; ~2,000-3,000
  items would be needed at the observed effect size. n = 600's MDE is ~0.025 on step F1.
- **Retrieval: per-item data survives** for all 33 pool-sweep cells, so the fallback ("aggregates
  only") does not apply. **15 matched pairs** (n = 750 each) differ only in retrieval. +CE raises
  ROUGE-L F1 in **15/15** pairs (8 significant uncorrected, 5 surviving Holm within the family) and
  **never moves exact match** (0/15 by any test, signs split 7/8).
- Multiplicity is reported honestly: under Holm, **no Task-A test survives** (smallest adjusted
  p = 0.1676), which is a caveat the masking-default decision should see.

## Ranked recommendation

Six ranked takeaways are in §7 of the note. The top three: (1) do not justify a masking default with
the composite; (2) typed > raw is defensible on step-level metrics but marginal and fragile under
correction; (3) typed vs uniform should be recorded as undecided at this n rather than given a
winner. Choosing among them is Jahid's call with his supervisor.

Unmeasured and flagged as such: pool size (all 15 pairs hold size fixed — the other half of #6
item 4), balanced vs imbalanced, any guided condition, MetaQA, any v2 run.

### DIFF
--- a/docs/adr/0005-seed-before-sampling-v1-sampled-results-not-reproducible.md
+++ b/docs/adr/0005-seed-before-sampling-v1-sampled-results-not-reproducible.md
@@ -3,6 +3,8 @@
 - **Status**: Accepted
 - **Date**: 2026-08-12
 
+*(Amended 2026-08-20 by [0020](./0020-prior-work-re-analysis-convention.md): v1 runs remain non-reproducible, but their surviving per-item artifacts may be re-analyzed as prior work under that ADR's conditions.)*
+
 ## Context
 
 During the v1→v2 port (PR #7, issue #3), the ml-engineer found and the independent review confirmed a v1 reproducibility bug: v1's router runners called `random.sample` on their evaluation questions **before** calling `set_seed`. All ten per-model copies of `router.py` carry the bug (verified in review); line numbers cited from `components/router/models/qwen2_5_0_5b/router.py` — sampling at lines 181–183, `set_seed` at line 217. Any v1 router result produced with `--sample_size` therefore used an unseeded draw — the sampled subset was never reproducible, in v1 or anywhere else. v2's operating contract (CLAUDE.md, Reproducibility) requires fixed seeds threaded through all randomness, and ADR [0001](./0001-v1-to-v2-migration-scope-and-method.md) governs how v1 results may be compared against.
--- a/docs/adr/0011-comparison-artifact-conventions-and-the-significance-claim-floor.md
+++ b/docs/adr/0011-comparison-artifact-conventions-and-the-significance-claim-floor.md
@@ -5,6 +5,8 @@
 
 Amends [0009](./0009-paired-bootstrap-and-mcnemar-as-the-significance-protocol.md) (the significance protocol of record) with three conventions established by PR #22 (issue #20 hardening). PR #22's Gate-1 review found the conventions sound but unrecorded; this ADR is that record. None of these came from the supervisor.
 
+*(Amended 2026-08-20 by [0020](./0020-prior-work-re-analysis-convention.md): analysis notes reporting a statistics battery carry a machine-readable JSON companion beside the note.)*
+
 ## Context
 
 PR #22 hardened `--compare` in `scripts/musique_decompositions_evaluator.py`. Three of its choices are protocol-shaped — they will appear in every run's config snapshot and every comparison output, and in November they would read as settled decisions unless their provenance is recorded.
--- a/docs/adr/0020-prior-work-re-analysis-convention.md
+++ b/docs/adr/0020-prior-work-re-analysis-convention.md
@@ -1,0 +1,52 @@
+# 0020. Prior-Work Re-Analysis Convention
+
+- **Status**: Accepted
+- **Date**: 2026-08-20
+
+Amends [0005](./0005-seed-before-sampling-v1-sampled-results-not-reproducible.md) (v1
+results stay non-reproducible as *runs*; their surviving per-item artifacts may still be
+re-analyzed) and [0011](./0011-comparison-artifact-conventions-and-the-significance-claim-floor.md)
+(analysis notes reporting a statistics battery carry a machine-readable companion).
+
+## Context
+
+The first re-analysis of v1 artifacts under v2's statistical protocol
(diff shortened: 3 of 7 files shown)


AhmadiJahid:
Fix pass for the delta review is at `75746a2` (branch tip). Not merged.

**I-1 — alignment traceability.** §3 now states the Task-A alignment key and order explicitly
(normalized question text, **sorted** — matching v2's `--compare`, which sorts its aligned ids so the
bootstrap is reproducible), stated next to Task B's positional-alignment statement. Added a measured
caveat: ran the typed-vs-raw battery in **both** row orders and observed point estimates agreeing to
**<= 2.8e-17** and **every verdict identical**, while CI endpoints move in the **third or fourth**
decimal — step F1 `[+0.0053, +0.0412]` sorted vs `[+0.0054, +0.0414]` file order (your example
reproduces), and exact match `[+0.0033, +0.0400]` vs `[+0.0050, +0.0400]`. Both orders are recorded
in the JSON under `task_a_masking.row_order_sensitivity`; the note tells the reader to take CI bounds
to three significant figures, not the last digit.

**I-2 — machine-readable statistics.** Added
`docs/analysis/2026-08-20-v1-masking-and-retrieval-significance.json` (statistics only; verified no
question or step text in any of its 417 strings). Per pair and per metric, overall and per gold hop:
mean difference, bootstrap CI + verdict, t/dof/p, McNemar counts+p on the binary metrics, composite
terms, the composite-difference decomposition, power (overall rows), Holm-adjusted p's for every
stated family, a CI-vs-t-test verdict-agreement audit, plus **input sha256 + mtime for all 36 v1
per-item files**, the alignment order, and seed / resamples / chunk / weights. Regenerated
in-session; **all note tables reconcile against it with 0 mismatches** (overall, per-hop and power
tables checked cell by cell, plus the Holm survivor lists). The caveat block now says outright that
the harness is uncommitted, that the figures were **verified by independent recomputation in this
review**, and that they are not re-derivable from committed code until the 6(b) shim lands. The shim
stays a follow-up.

**Nits** — all applied: `R5` -> `6(b)`; the reference-term-removed diagnostic prints its renormalized
weights (**0.5 / 0.375 / 0.125**) and keeps its "not a house metric" label; §7 items 1 and 3 recast
to conditionals ("a typed default justified by the composite would rest on an artifact" / "the data
cannot name a winner between typed and uniform at n = 600"); n-needed ceiled (**1864 / 3140 / 2052 /
2138**); `docs/prior-work.md` line 80 now reads "shows no significant effect on exact match (0/15 by
any test, signs split 7/8)" — and I applied the same wording to the note's own §5 and §7.4, which
carried the identical overclaim.

**One harness bug found while building the audit** (worth your eye): a stratum where every paired
difference is exactly 0 — both systems score 0 exact match on every item of that hop — left the
t-test verdict unset, and the audit read `False != None` as a CI-vs-t disagreement. Those 5 Task-B
per-hop cells are *decided not-significant*, not undecided (CI `[0,0]`, t undefined with p=1.0,
McNemar 0/0 discordant with p=1.0). Fixed, and the audit now reports what the note claims: Task A
**15/0** overall and **60/1** with per-hop; Task B **75/0** overall and **300/2** with per-hop, the
two being `size1000_imbalanced__uniform` hop 3 hop EM (CI `[+0.0000, +0.1520]`, t p=0.0486) and
`size4000_imbalanced__typed` hop 2 ordered accuracy (CI `[+0.0003, +0.0683]`, t p=0.0502). No
reported statistic changed.

PR #36, opened 2026-08-20
Paired t-test in the significance battery + a v1-format --compare shim

### LINKED ISSUE #30: Report a paired t-test alongside bootstrap CIs + McNemar

AhmadiJahid:
Two related additions to the comparison path, both in `scripts/musique_decompositions_evaluator.py` `--compare` (which is where the whole ADR 0009 battery lives — there is no separate stats module; `scripts/compare_decomposer_arms.py` shells out to `--compare` and only reads its output).

## 1. Paired t-test alongside bootstrap CIs + McNemar (issue #30)

Two-sided `scipy.stats.ttest_rel` over the **same per-item differences, same pairing, same aligned items** the bootstrap resamples. Each row reports `t_statistic`, `degrees_of_freedom` (n − 1) and `p_value`, with `significant = p < alpha` and the ADR 0011 `underpowered` flag applied on the same terms as the other families.

- **Which metrics.** The five compared metrics that *have* a per-item value: `rouge_l_f1`, `step_f1`, `ordered_step_accuracy`, `exact_match`, `hop_count_exact_match` (`T_TEST_STATISTICS`, derived from the two existing constants). `composite_score` has no per-item value — its reference term is a micro rate, its step-count term a MAE — so it keeps only its bootstrap CI. This is exactly how the committed 2026-08-20 analysis note applied it, EM included.
- **Additive, never a replacement.** Bootstrap + McNemar remain the headline protocol (ADR 0009, as amended by ADR 0017 item 4). A new `tests_reported` block records the counts (4 bootstrap + 2 McNemar headline, 5 t-tests alongside) instead of the old hardcoded "six tests" sentence, and the note states that none is corrected for multiple comparisons and that the t-test rows re-test the same metrics on the same items rather than adding independent tests. The normality caveat that motivated ADR 0009's choice is printed next to the numbers.
- **Degenerate rows make no claim.** n < 2, or a zero sd of the per-item differences (e.g. a file compared with itself): `t_statistic: null`, `p_value: null`, `significant: false`, plus a `degenerate` reason string. scipy returns NaN/±inf there and neither is JSON-representable.
- **Dependency posture.** `scipy==1.17.1` is already pinned in `requirements.txt` and is what the analysis note used, so the p-value comes from it rather than a hand-rolled tail. Nothing was added to `requirements.txt`. (The hand-rolled McNemar stays hand-rolled — ADR 0009 wants exact integer arithmetic there, which is a different reason from dependency avoidance.)
- `compare_decomposer_arms.py` surfaces the new block in its table, read with `.get` so an older comparison artifact without it does not break the script.

## 2. The v1-format `--compare` shim (`--v1-per-item`, ADR 0020 / analysis note §7.6(b))

`--compare` can now consume v1's bare-list per-item files (same per-item field names, no `item_id`, no stamped composite weights).

- **(a) Explicit, never silent.** Opt-in flag. Without it a v1 file is still refused, and the error now names the flag. With the flag, a v2 object aborts too, and mixing one of each is explicitly unsupported (the two were scored by different code).
- **(b) Provenance recorded.** A `v1_format_inputs` block carries the prior-work caveat (**no commit SHA**, Gate 2 unsatisfied, citable prior work and not a v2 measurement), each input's sha256 / mtime / row count (ADR 0020 condition 2), the alignment and its definition, and the fact that the composite weights are the config's because v1 stamps none — `config_weights_match_per_item_files` is then `null`, not a bool that would read as "checked and equal". The caveat **leads** the run note and the stdout (condition 5).
- **(c) v1 stays read-only.** Only reads, plus a source-level guard (ADR 0016 style): a `--run-dir` inside `paired_comparison.v1_compat.read_only_prior_work_root` is refused before any work happens.
- **(d) Alignment is statable.** `--v1-alignment normalized_question` (default from config; key = normalized question text, rows in sorted order of it — the alignment the analysis note states for Task A) or `position` (row i with row i, needed for the note's Task B where one question text appears twice). Under either, every pair's `question` and `gold_steps` are asserted equal, so a misordered file aborts instead of quietly comparing different items.

## Ran-X-observed-Y

| check | command | observed |
|---|---|---|
| unit tests (file) | `.venv/bin/python -m unittest tests.test_decomposition_metrics` | 19 → **40** tests, OK |
| full suite | `.venv/bin/python -m unittest discover -s tests` | **235 tests, OK** |
| full-pipeline smoke | `.venv/bin/python scripts/smoke_test.py` | **36/36 stages passed** |
| t-test known answers | closed-form Student-t survival at dof 1 and 2 | t = 1.0 → p = 0.5; t = 0.5 → p = 2/3; t = 29 → p = 1 − 29/√843 — matched to 9 places |

**Shim acceptance check** — the shim run against the committed `docs/analysis/2026-08-20-v1-masking-and-retrieval-significance.json`, seed 42, the note's stated alignment:

| pair | n | compared values | result |
|---|---|---|---|
| Task A typed vs raw (`normalized_question`) | 600 | 37 | **all bit-identical, worst delta 0.0** |
| Task A uniform vs raw (`normalized_question`) | 600 | 36 | all bit-identical, worst delta 0.0 |
| Task B `size1000_balanced_trial0__raw` (`position`) | 750 | 36 | all bit-identical, worst delta 0.0 |

Bit-identical includes the four house **bootstrap CI edges** (so the alignment matches the note's), both **McNemar** p-values and b/c counts, and all five **t statistics and p-values**. The recorded input sha256s and `mtime_utc` values match the note's `inputs` block exactly. v1 was not modified: `git status` clean in `/cta/users/fyilmaz/Thesis---QA` at HEAD `a020fd6`, input mtimes unchanged.

## Decisions for the lead

1. **Two columns of the analysis note remain non-reproducible by design.** The note bootstraps `exact_match` and `hop_count_exact_match` (and a `composite_no_ref_renorm` diagnostic); v2's `BOOTSTRAP_STATISTICS` does not, and adding them would change ADR 0009's protocol (its "six tests"), so I did not. Everything ADR 0009 covers reproduces exactly. Whether to widen the bootstrap is Jahid's call.
2. **The comparison output is not schema-versioned.** ADR 0011 convention 2 versions the *per-item* file (`musique_decomposition_per_item/1`); the comparison metrics JSON carries no `schema` key on `main`, so there was nothing to bump. New keys are additive (`t_test`, `tests_reported`, `v1_format_inputs`), and `config_weights_match_per_item_files` can now be `null` on the v1 path only. Introducing a version for this artifact would be a new convention and wants an ADR — flagging rather than doing it.
3. **No ADR written.** The t-test decision is already ADR 0017 item 4 and the shim is already ADR 0020's named open gap; the implementation conventions (which metrics get a t-test, degenerate-row handling, the two alignments and the same-item assertion) are documented in `docs/METRICS.md` §5. If they should live in the ADR trail, amendments to 0009/0011/0020 are the natural home — that is a Jahid/supervisor-shaped call, not mine.

Closes #30
Refs #6


AhmadiJahid:
## Gate-1 fixes pushed — `2d8bf49` (rebased onto `origin/main` `d161ffc`)

Branch history is now `078e551` (original feature commit, rebased) + `2d8bf49` (this fix pass). No merge from me.

### I-1 — the same-item verification can no longer be vacuous

The witness fields are now **required, not opportunistic**. The reviewer's case reproduced and now refuses:

```
$ ... --compare vacuous_a.json vacuous_b.json --v1-per-item --v1-alignment position
--v1-per-item --v1-alignment position: refusing to compare, because nothing can verify
that the paired rows are the same evaluation item.
  required on every row of both files: ['question', 'gold_steps']
  missing: ['question', 'gold_steps']
... (ADR 0020 condition 3). rc=1, no metrics file written
```

I went **stricter than the finding asked**, for one reason worth flagging: PR #35 landed on `main` while I worked and amended ADR 0020 condition 3(b) to require, for positional alignment, that *the alignment field **and** the gold* be asserted equal on every matched row. "At least one witness" would have left the code laxer than the convention it implements, so:

| alignment | required witnesses | why |
|---|---|---|
| `position` | `question` **and** `gold_steps` | ADR 0020 condition 3(b) as amended by PR #35 |
| `normalized_question` | `gold_steps` | the question is the alignment key — equality holds by construction and witnesses nothing |

Four negative/positive controls added: witness-less (both missing) refuses; question-only under `normalized_question` refuses naming "true by construction"; gold-only under `position` refuses citing condition 3(b); question+gold passes and records `verification_fields` with 40/40 pairs. The record now carries `verification_fields`, `fields_not_usable_as_witness` and `tautological_fields`, and `_v1_verification_sentence` builds the note line from the counts — a separate test pins that zero and partial counts render as `NOT VERIFIED`.

### I-2 / I-3 — dated records, no silent rewrites

- **ADR 0020 Consequences**: dated note stating what is re-derivable (the ADR 0009 battery over a pair's full item set) and the five families that stay harness-only — EM/hop-EM bootstrap CIs, `composite_no_ref_renorm`, the `power` blocks, `holm_bonferroni`, `per_gold_hop` strata — each with its reason, pointing at `docs/METRICS.md` §5 for the shim conventions. Also records that condition 3 is now *enforced in code* on that path, not merely reported.
- **Masking note**: dated bracketed addenda at both claim sites the review named (the caveat block's "not re-derivable until the shim lands", and recommendation 6(b)'s "every number"), plus §1's reproducibility-gap paragraph, which carried the same claim. The dated analysis record itself is untouched otherwise.
- **ADR 0011**: dated note at conventions 2/3 — the additive keys, `config_weights_match_per_item_files` now tri-state, the bare-list refusal still standing off the opt-in path, and the comparison artifact remaining **unversioned** with that as an open Jahid call. Record only; no `schema` key added.

### Nits

- **N1** guard hoisted out of the `if args.v1_per_item:` branch — unconditional now. Pinned by an **AST test** (the house pattern from `tests/test_generation_contract.py`) rather than by aiming a real `--run-dir` at the read-only tree: a regression in *that* test would itself write into the tree the rule protects.
- **N2** `_require_matching_weights` no longer called on the v1 path, with a comment saying why and pointing at `composite_score_weights_source`.
- **N3** all three v1 refusal paths (id mismatch, duplicate key, same-item mismatch) now print `a: row 4, b: row 4 (key sha256:87e754f49fda)`. Verified with a fabricated question string: absent from every message, `sha256:` and `row N` present.
- **N4** comment fixed.
- **N5** docstring corrected (NaN/Infinity *are* written by the json module — they are simply not valid JSON, which is the actual problem). The non-finite guard covers t, p **and the row's three means** — a NaN input made `system_a`/`difference` NaN while t/p were already nulled, so `json.dumps(row, allow_nan=False)` still failed; that assertion is now a test. And a v1 row with a `null`/non-finite metric exits with `row 7 has non-numeric or non-finite compared field(s): ['step_f1=None']` instead of a `TypeError`.

### Ran at the fix SHA (`2d8bf49`, post-rebase)

| check | observed |
|---|---|
| `tests/test_decomposition_metrics.py` | 40 → **49** tests, OK |
| full suite | **272 tests, OK** |
| `scripts/smoke_test.py` | **36/36 stages passed** |
| acceptance reproduction (3 pairs) | **129/129 values bit-identical**, unchanged by every fix above |
| v1 tree | clean at `a020fd6`, 0 modified files |

The acceptance figure is 43 values per pair × 3 pairs (I added `min_attainable_p_value` and `dof` to the comparison set since the first run's 37/36/36).

Issue #27, opened 2026-08-20
Few-shot-prompted router + with/without-router MuSiQue evaluation


AhmadiJahid:
From the 2026-08-12 meeting (unrecorded idea, surfaced by the transcript cross-check, triaged by Jahid 2026-08-20 on #19): evaluate a few-shot-prompted router, and run a with-router vs without-router comparison on the MuSiQue eval set (ADR 0007).

**Outcome.** A measured answer to "does the router earn its place": the same decomposition pipeline evaluated with and without router-predicted hop counts, plus a few-shot-prompted router as a candidate implementation.

**Sequencing.** After #12's numbers exist (they say how much hop guidance matters at all) and alongside the #23 regressor design — this may become #23's evaluation harness.

Refs #23, #12, #19, ADR 0010.

AhmadiJahid:
**Lead's drift check on exp-012's headline comparison** (done independently while the run was in flight; recorded so the log row can cite it and a reviewer can re-derive it).

exp-012 compares a **freshly generated** `router_guided` arm (at `c1369a9`) against exp-004's `unguided` arm, which was **generated at `6fc4bba`** and is only re-scored at the current commit. The evaluator is therefore held constant, but the *generation* is not — so the question is whether the unguided generation path drifted between those commits. It did not:

- `components/decomposer/run_decomposer.py`: +283/-9 between `6fc4bba` and HEAD, and every hunk is the issue #27 `router_guided` addition — `hop_source` added to `_CONDITION_KEYS`, the new `resolve_hop_source`, and `load_predicted_hops`. `resolve_hop_source` refuses `predictions` in an unguided arm and short-circuits on `if not guided:`, so an unguided run never enters the new code.
- `configs/decomposer_musique.json`: +14/-1, all additive — a `hop_source: "gold"` default, the `hop_predictions` block, the `router_guided` condition, and the `_conditions_note` rewording. **The `unguided` condition block is untouched** and `generation_overrides.max_new_tokens` stays 1024.
- `components/decomposer/models/mistral_7b_instruct/` (the prompt files): **empty diff**. The unguided prompt still hashes to exp-004's recorded `e9fef279…`.
- Retrieval artifact: exp-012's run log records `input_sha256 e5c418a9b25f5ef290a78dd3a99642e4b0c5d7b6fbe6b5e2785a1424239ae70d`, **byte-identical** to the artifact recorded in `experiments/exp-004/config.json`.
[...]

Issue #15, opened 2026-08-17
Hop-matched retrieval: measure the assumption (mixed vs oracle-hop-matched vs router-hop-matched)


AhmadiJahid:
**Outcome.** Measure whether retrieving few-shot examples of the same hop depth as the query actually helps: three retrieval conditions — mixed (no hop constraint), oracle-hop-matched (gold hop depth), router-hop-matched (predicted hop depth) — on the MuSiQue eval set.

**Why it matters.** "Hop-aware" is in the contribution statement (ADR 0006) as an assumption, not a measured effect. Oracle vs router also isolates how much of any gain survives an imperfect hop predictor, which feeds the router question (ADR 0006 item 6, #12).

**Done when.** All three conditions are evaluated on the same eval set (ADR 0007) with log entries, so the gap between mixed, oracle and router is a number rather than an assumption.

AhmadiJahid:
**Machinery built (PR #39, in Gate 1 review).** Hop-bucket candidate filter before top-k in `check_question_similarity.py`; hop source pluggable between `gold` (query id parse) and a predictions JSONL keyed by query_id. Off by default; feature-off retrieval verified byte-identical to pre-change behavior. Implementer defaults in ADR 0022. Once merged, the **mixed** and **oracle-hop-matched** conditions are runnable (CPU precompute now, decomposer generation when a GPU window opens; eval set is ADR 0007's pinned 600).

Two things surfaced during the build:

1. **Router-hop-matched is blocked on a predictions source.** The interface expects `{query_id → predicted_hop}` JSONL; v1's `run_router.py` emits a JSON array with no query id, so it does not conform (keying on question text was rejected — masking modes rewrite it). Where predictions come from (existing router re-run with ids, the ADR 0010 regressor, or a few-shot router per #27) is a Jahid + supervisor call, tied to #23/#27.

2. **A confound to state wherever #15's numbers are reported:** hop-matching shrinks the candidate set (the 4-hop bucket is ~1.2k of the ~9.2k pool), so any gain mixes "same hop depth" with "smaller candidate set". A size-matched mixed control would separate the two; not built — flagging it as a design option for Jahid rather than deciding it.

Refs PR #39, ADR 0022, #23, #27.

AhmadiJahid:
## Results — exp-009 (mixed vs oracle-hop-matched), 2 of this issue's 3 conditions

This issue's only prior comment predates the run ("once merged, the conditions are
[...]

PR #43, opened 2026-08-22
Router readiness: query-id-keyed predictions, few-shot-prompted router, router-guided decomposer arm

AhmadiJahid:
Refs #27 (the with/without-router evaluation run itself is a separate experiment lane, so this does not close it). Also unblocks the `predictions` hop source of #15 / ADR 0022 item 5, and serves ADR 0023 item 2.2.

**Nothing here is a measurement.** No router has been run: no accuracy number, no effect on decomposition quality, no log entry. Everything below was executed CPU-only with no weights loaded.

## What changed and why

**1. `components/router/run_router.py` — predictions keyed by query id.** A run whose question source carries ids (`questions_format: jsonl`) writes `predictions.jsonl`: one object per query with `query_id` / `predicted_hop` (names from config), plus `expected_hop`, `correct` and `parse_fallback`. That is exactly the shape ADR 0022 item 5 documented and could not find a producer for, so the two consumers join on the id with no adapter: the retrieval chain's `--hop-source predictions` and the decomposer's new `router_guided` arm. `detailed_results.json` now carries `query_id` too. A blank/absent or repeated id is refused *before* a model loads; `parse_fallback` and `unparsed_response_rows` are new because v1 returned `parsing.default_hop` silently, so a run's accuracy mixed predictions with defaults and nothing recorded which. A dry run writes no predictions file — it generates nothing, and fabricating routing decisions is not a smoke test; the file's shape is pinned instead by a unit test that round-trips it through `load_predicted_hops`.

**2. A retrieved-few-shot prompting mode + `configs/router_musique.json`.** Exemplars are the query's own candidates from the retrieval artifact the decomposer already reads (`--retrieval-input` / `retrieval.input_key`), each labelled with **its own** gold hop depth parsed from its pool id, with the query dropped from its own exemplars by id and by normalized question text — `run_decomposer.is_self_example`, imported rather than copied. Nothing is trained. New shared prompt `components/router/models/prompt_few_shot_musique.md` (the v1 per-model prompts are untouched; MuSiQue's 2/3/4 answer digits come from a `parsing_overrides` block, not an edit to a v1 asset). The config asserts the pinned ADR 0007 set by id, so a predictions file over a different 600 is refused rather than joined against arms it does not cover.

**3. `components/decomposer/run_decomposer.py` — the guided condition could only take gold hop counts; now it can take a router's.** `hop_source` joins `guided`/`stop_after_step_lines` as a condition key, and `router_guided` sets it to `predictions`: the hop count comes from a predictions JSONL joined on the query id (`--hop-predictions`) and overrides the gold depth even when they disagree. Nothing falls back — a query the file does not cover is refused with the offending ids named. `hop_count` in `results.json` keeps its meaning (the **gold** depth, so no existing consumer changes and a prediction cannot move a row into another hop's table); `prompt_hop_count` and `hop_count_source` are new, and `metrics.json` gains a `hop_source` block with the predictions path, its sha256 and the gold-agreement counts.

**4. Tests, fixtures, docs.** `tests/test_router_predictions.py` (44 tests), `tests/fixtures/router/hop_predictions_musique.jsonl` (9 hand-written rows over the fixture ids, 3 deliberately disagreeing with gold, 1 flagged as a defaulted hop — synthetic, no dataset rows), two new smoke stages, ADR 0024, forward notes on ADR 0022 items 5 and 9's consequence, README updates, and the ADR index gained the missing 0023 row plus 0024. `tests/test_decomposer_conditions.py` was updated for the fourth condition and the two new result fields.

## Evidence (ran-X-observed-Y, all CPU, no weights)

- `python -m unittest discover -s tests` → **Ran 354 tests, OK** (310 on `main` + 44 new).
- `python tests/test_decomposer_conditions.py --skip-data-checks` → **258 checks passed**.
- `python scripts/smoke_test.py` → **39/39 stages passed**, including `router_dry_run_musique_few_shot` and `decomposer_dry_run_musique_router_guided`.
- Few-shot router dry run on the fixtures (9 queries): 5 exemplars per prompt, each `A: <hop>` from its pool id, **2 self-examples dropped** (the two planted fixture cases), `predictions_file: null` with the "dry run predicts nothing" note.
- `router_guided` vs `oracle_guided` dry runs on the same 9 fixture queries: the routed arm put the **predicted** hop in all 9 prompts (row-by-row equal to the predictions file, 6 agreeing with gold / 3 disagreeing), the oracle arm put the gold one, same query ids in the same order, and the routed run's per-hop counts stayed 3/3/3 on the **gold** depth.
- Refusals exercised end to end: a predictions file missing one query (names it), `--hop-predictions` on the oracle arm, a few-shot config with no exemplar source, and a few-shot config pointed at a v1 prompt with no `{few_shot_examples}`.
- Parameter ceiling: **not asserted anywhere in this verification** — every run was a dry run, so no model was loaded (the ADR 0016 caveat). The router's assertion path (`assert_within_ceiling`, `router_max_params` = 8e9) is unchanged.

## For the lead to decide

1. **A zero-shot MuSiQue router arm is not included.** `configs/router_musique.json` requires the few-shot prompt; running it with `prompt_zero_shot.md` is refused rather than silently prompting zero-shot under a few-shot label. A zero-shot-on-MuSiQue comparison arm would be one more config — not in the brief, so not added.
2. **`few_shot.exemplar_hop_source` has one implemented value (`pool_id`).** Labelling exemplars by their gold *step count* instead (what the decomposer's exemplar hop lines state, ADR 0013) is a different quantity and a research choice; refused rather than guessed.
3. **The router component now imports `run_decomposer.py`** for the self-exclusion rule, the question reader and the pinned-set assertion (`run_answerer.py` already does this for the last). Consolidating those three into `src/` is the obvious cleanup and is deliberately deferred — same reasoning as ADR 0022 item 6.
4. ADR 0024 is agent-authored implementer design, **pending Jahid**.

### DIFF
--- a/components/decomposer/README.md
+++ b/components/decomposer/README.md
@@ -14,7 +14,8 @@ placeholders that refer to earlier answers. Ported from v1, with the four per-mo
   few-shot and post-processing settings.
 - `configs/decomposer.json` — MetaQA: seed, guided flag, embedding-model registry, retrieval defaults.
 - `configs/decomposer_musique.json` — the MuSiQue variant (issue #12): the pinned 600-question
-  evaluation set of ADR 0007, hops 2/3/4, and the three run conditions.
+  evaluation set of ADR 0007, hops 2/3/4, and the run conditions (including `router_guided`,
+  issue #27's with-router arm).
 
 ```bash
 # few-shot examples from a reranked/truncated retrieval file
@@ -34,15 +35,35 @@ python components/decomposer/run_decomposer.py --model mistral_7b_instruct \
 ## Run conditions (`configs/decomposer_musique.json`)
 
 `--condition` selects a named block from the config's `conditions`; the config's
-`condition` key is the default when the flag is omitted. A condition may set only `guided`
-and `stop_after_step_lines` — model, seed and decoding are shared, and the runner rejects a
-condition that tries to move them, so the arms cannot drift apart.
+`condition` key is the default when the flag is omitted. A condition may set only `guided`,
+`hop_source` and `stop_after_step_lines` — model, seed and decoding are shared, and the runner
+rejects a condition that tries to move them, so the arms cannot drift apart.
 
 | condition | prompt | generation |
 |---|---|---|
 | `unguided` | no hop count | `generation_overrides.max_new_tokens` only |
 | `oracle_guided` | the query's gold hop count, and each exemplar's own gold hop count | same |
+| `router_guided` | the **router's predicted** hop count for that query id, exemplars unchanged | same |
 | `unguided_capped` | no hop count | stops after `stop_after_step_lines` step lines (default 8) |
+
+**The with-router arm (`router_guided`, issue #27).** Identical to `oracle_guided` except where
+the query's hop count comes from: a router predictions JSONL (`--hop-predictions`, written by
+`components/router/run_router.py`), joined on the **query id**, whose prediction overrides the
+gold depth even when the two disagree — that disagreement is the behaviour under test. Nothing
+falls back: a query with no prediction in the file is refused, naming the offending ids, because
+filling it in from the gold depth would make the arm a blend of two conditions. The per-hop
+reporting and the pinned-set assertion stay on the **gold** depth, so a prediction cannot move a
+row into another hop's table: `results.json` carries `hop_count` (gold, unchanged meaning),
+`prompt_hop_count` (what the prompt stated) and `hop_count_source`, and `metrics.json`'s
+`hop_source` block carries the predictions path, its sha256 and how often the prediction agreed
+with the gold depth.
+
+```bash
+python components/decomposer/run_decomposer.py --model mistral_7b_instruct \
+    --config decomposer_musique.json --condition router_guided \
+    --hop-predictions runs/router_musique/few_shot/<run_id>/predictions.jsonl \
+    --retrieval-input <the pinned top-5 JSONL over the 600 questions of ADR 0007>
+```
 
 The cap is a `StoppingCriteria` during decoding, plus a trim of the partial line the
 criterion can leave behind once it fires. Step lines are counted **after** `<think>`/tail
--- a/components/decomposer/run_decomposer.py
+++ b/components/decomposer/run_decomposer.py
@@ -250,11 +250,14 @@ def resolve_hop_source(
     where the number came from and in nothing else, which is what makes a with-router versus
     without-router comparison a comparison.
 
-    Four refusals, all of them a run that would otherwise be filed under the wrong label:
+    Five refusals, all of them a run that would otherwise be filed under the wrong label:
 
-    - an unimplemented hop source (not silently treated as ``gold``);
+    - an unimplemented hop source (not silently treated as ``gold``), including a condition
+      that sets ``hop_source`` to null;
     - ``predictions`` in an **unguided** arm - no hop count reaches the prompt at all there,
       so the run would be the unguided arm wearing the router's name;
+    - ``predictions`` with ``few_shot_exemplar_hop_count`` ``"query"``, which would stamp the
+      *prediction* on every exemplar (ADR 0013's defect, with a prediction for a gold depth);
     - ``predictions`` with no file named;
     - ``--hop-predictions`` passed to a ``gold`` arm - the file would be ignored, and the
       run recorded as the oracle while the operator believed it routed.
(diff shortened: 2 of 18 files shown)


AhmadiJahid:
## Gate 1 fix pass @ `6247211` — all 10 findings addressed

Verified CPU-only, no weights: **372 tests OK** (was 354; this file 44 → 62), **smoke 39/39**, `tests/test_decomposer_conditions.py --skip-data-checks` **258 checks passed**.

**C1 — parsing preferred a regurgitated exemplar answer.** Confirmed and fixed three ways, in order of what actually guarantees the read:
1. `response_truncate_at` (new required key, per config because it is a property of the *prompt shape*) cuts the response at the answer before any hop count is read — `["\n"]` for the retrieved-few-shot prompt, `[]` for the v1 MetaQA prompts whose responses are multi-line by design, so their parsing is untouched. Implemented as `clean_response`, which reuses `src/step_lines.py`'s think-stripping and then truncates; the order matters and is commented (that helper truncates before stripping outer whitespace, so a leading newline would otherwise cut the answer away — there is a test for exactly that).
2. `router_musique.json` now sets `parsing_overrides.answer_regex: null`. The prompt consumes the `A:` cue, so a rule requiring an `A:` prefix can only match regurgitation; the next rule reads the leading digit and builds its digit class from `hops`, so no hop-specific constant is left in the config.
3. `generate()` passes `eos_token_id` (as the decomposer does) so a finished model can stop. Stated as *not* the guarantee: a model can always keep going, which is why truncation is the fix.

Measured on the committed configs: `"2\n\nQ: Who founded the press in Marlow Bay?\nA: 3\n\nQ:"` → **(3, True)** before, **(2, True)** now. `TestResponseParsing` covers the regurgitation case for every hop pair, plus the leading-newline, `<think>`-preamble and v1-multi-line cases. You were right that the old test validated a shape the prompt cannot produce — `test_the_exemplar_answer_line_is_readable_by_the_run_s_parser` (which asserted an `A: <n>` *response*) is gone, replaced by checks written against the shape the prompt actually elicits, and ADR 0024's verification section now says so explicitly rather than quietly restating a corrected number.

**I1 — MetaQA-only fallback tables.** `_NUMBER_WORDS` covers 1–9 and `hop_word_labels(hop)` derives the labels from the depth, so the fallback follows the config's `hops` (`"four hops"` → `(4, True)`, was `(2, False)`). New `assert_parsing_covers_hops`, called at config-load time and therefore exercised by every dry run, refuses: a depth the fallback table cannot name, a depth above 9 (the digit class is a single-character regex class), and `default_hop ∉ hops`. Its record ships in the snapshot and metrics as `parsing_coverage`. Defaults are broken out as `unparsed_response_rows_per_gold_hop` and in the run note, because a defaulted row is scored as a prediction and so flatters `default_hop`'s own class.

**I2 — accuracy reported without definitions.** `ACCURACY_DEFINITIONS` sits in `metrics.json` next to the numbers: gold = the hop parsed from the query id (never a prediction), `overall_accuracy` = exact-hop **including** defaulted rows, `hop_<h>_accuracy` = **within-gold-class recall, not precision** (with the reason it is easy to misread), and what `unparsed_response_rows` does to those numbers. The run note repeats the one-liner.

**I3 — missing fifth refusal.** `hop_source: predictions` + `few_shot_exemplar_hop_count: "query"` is now refused in `resolve_hop_source`, naming ADR 0013 and pointing at `exemplar_gold`. Refused for the config combination regardless of whether few-shot is on for a given run, so it cannot reach a prompt at all; a test also pins that the committed config uses `exemplar_gold`, which is what makes this a guard and not a fix.

**Nits.** N1 post-filter self-exclusion assertion added, mirroring `run_decomposer.py:653-662`. N2 `output_subdir_zero_shot` removed from `router_musique.json`; its absence is a named refusal (`optional` + a message pointing at the key and `--output-root`) rather than a `require()` KeyError, with a CLI test. N3 `sys.path` now inserts the decomposer dir first and `src` second, so `src` stays ahead. N4 an explicit `"hop_source": null` in a condition is refused instead of inheriting the default. N5 no action, as agreed. N6 README states the `predictions.enabled` gate, the `lines`-source reason it is off for MetaQA, and that a dry run writes no file — plus a new paragraph on the truncation/coverage/default-scoring behaviour.

**Convention capture.** ADR 0024 gains **item 1a** (implementer design, pending Jahid): the truncation rule and why it is per-config, the parsing-coverage invariant now guarded in source with ADR 0016's reasoning for why a `_note` was not enough, and a plain statement of **how a defaulted hop is scored** — recorded as a prediction, counted in the accuracies, denominator unchanged, hence the per-gold-hop split, and the note that whether to exclude or refuse such rows instead is Jahid's call. Item 8 now lists five refusals; a new consequence records that a new prompt for this component owes a `response_truncate_at` and a response-shape test.

No merge. Ready for the delta-confirm.

AhmadiJahid:
Gate 1: full review at ca1d5a4 (BLOCKED: C1 + I1-I3 + 6 nits) → fix pass 6247211 → delta-confirm: **CONFIRM, merge unblocked**. All blocking findings verified resolved by independent probes (C1 probe now (2, True) on all 10 model folders; v1 MetaQA parsing parity 0 diffs across 20 probes × 10 folders).

Residual nits recorded for a future pass (none blocking): (1) refusal-count docstring says three, has five raise paths; (2) uncovered-label refusal unreachable under hop_words; (3) eos_token_id docstring states a model-dependent equivalence as fact; (4) ACCURACY_DEFINITIONS omits the --num-runs>1 mean/std keys; (5) a single-line UNCLOSED <think> block can still yield a confident parse — router should consult src/step_lines.has_open_think_block like the decomposer does; (6) bare number word without "hop" still defaults (v1 semantics preserved, now visible per gold hop).

Reminder recorded in the review: ADR 0024 is implementer design pending Jahid — the truncation rule and "a defaulted hop is scored as a prediction" stance need his confirmation before any router number is reported from them.

Issue #40, opened 2026-08-22
Composite score's reference_validity term is decided by 2/600 items: _REF_RX matches [#k], the eval set uses bare #k


AhmadiJahid:
Surfaced by the 2026-08-22 error analysis (docs/analysis/2026-08-22-finetuned-vs-prompting-error-analysis.md, commit ec2f844), which reproduced the published exp-008 bootstrap bit-identically before decomposing it.

**Measured, on the ADR 0007 pinned 600:**
- `_REF_RX = r"\[#(\d+)\]"` in `scripts/musique_decompositions_evaluator.py` matches only bracketed `[#k]` references. The eval set's gold uses bare `#k` exclusively (1200 references, 0 bracketed).
- The composite's 0.2-weight reference_validity term therefore has an empty denominator (scored 1.0 by rule) on five of seven evaluated arms, and 0.0 on exp-004 unguided/unguided_capped — where exactly two items own the entire micro denominator.
- This artifact accounts for 95.3% of the Mistral unguided→oracle_guided composite doubling (issue #12) and 93% of the apparent Qwen-over-Mistral composite gap, and it is the mechanism behind the bimodal/asymmetric bootstrap CIs on composite in experiments/exp-008/metrics.json.
- PR #38 §4.3's prediction (composite can rank junk above real systems) has now fired in committed v2 numbers: exp-004 unguided composite 0.2098 vs question-echo 0.2333 / three-junk-steps 0.2778.

**Not decided here (Jahid + supervisor):**
1. Is `_REF_RX`'s bracketed syntax a defect or an intended v1-carried definition?
2. The primary-metric question (issue #6 item 5): the exp-008 headline result is unaffected on step_f1 / ordered_step_accuracy (tight, symmetric CIs, full significance battery) — whether those lead and composite is dropped/repaired is the open decision. The analysis note §8 ranks the options without choosing.

No code change proposed until the humans decide; any evaluator change is shared-pipeline and goes through Gate 1, and re-scoring is cheap (per-item outputs are on disk).

AhmadiJahid:
Jahid authorized work on this (2026-08-22, ADR 0023 item 1): a literature-grounded replacement/repair for the composite is now an active lane — analyst survey first, Gate-1-reviewed implementation after, existing per-item outputs re-scored rather than re-run. Constraint unchanged: no closed commercial model judges decomposition quality. Primary-metric adoption remains supervisor-gated (issue #6 item 5).

[...]

PR #44, opened 2026-08-22
Break-faithful EM/SARI/GED + repaired chain validity as additive evaluator metrics (Refs #40)

AhmadiJahid:
Implements the model-free, per-item, re-scorable part of the shortlist in `docs/analysis/2026-08-22-metric-candidates.md` §4 (commissioned by ADR 0023 item 1). **Additive only**: new columns beside the existing ones. `composite_score`, `reference_validity_*` and `_REF_RX` are untouched, so **no committed number changes value**, and **no metric is adopted** — the seven §6 decisions (which metric leads, the composite's fate, whether the no-closed-model constraint extends to open-weight scorers) stay with Jahid and his supervisor. Refs #40; adoption stays open.

## What it adds

Four per-item columns, their macro aggregates, their per-gold-hop breakdowns, and their registration in the ADR 0009 battery:

| column | source | direction |
|---|---|---|
| `break_exact_match` | official Break `get_exact_match` (lowercased `@@SEP@@`-joined string, no punctuation stripping) | higher |
| `sari` | official Break `sari_hook.py` (keep-F1 + add-F1 + delete-precision)/3, n = 1..4, source = the question | higher |
| `ged` | official Break `normalized_graph_edit_distance` over the reference graph | **lower — it is a distance** |
| `chain_validity` | house repair (PR #38 §5 option 3(a)): per item, bare `#k`, **no free credit** for a prediction that emits no reference where the gold chains | higher |

`norm_EM` is deliberately **not** ported: its normalizer is ~14 QDMR-operation-specific rules over spaCy parses plus a 16-way operation classifier (survey §1.3).

Battery: `BOOTSTRAP_STATISTICS` / `MCNEMAR_STATISTICS` / `T_TEST_STATISTICS` were closed tuples; they now cover the new metrics (7 bootstrap + 3 McNemar + 9 t-tests on v2 inputs, counts in `tests_reported`). Because `ged` is the only distance in the report, every bootstrap/McNemar/t-test row carries `direction` and, when significant, `favours` (direction already applied), the metrics JSON lists `lower_is_better_statistics`, and the run note's table has a `better` column with each verdict annotated `yes (favours a|b)` — a negative difference on `ged` cannot be read as a loss.

## The three GED policy calls (survey §5 item 4), and their justification

Recorded in ADR 0026, marked **implementer design, pending Jahid**:

1. **Substitution cost / spaCy.** Break's cost is `1 - edit_distance.SequenceMatcher.ratio()` over spaCy `en_core_web_sm` *lemmas*. spaCy is not a dependency and its model needs a download, and the evaluator must stay runnable offline — so the ratio and its match-counting DP are ported faithfully (including the tie-break order; matches on a minimum-cost path is **not** the LCS length) over lowercased whitespace tokens. **Consequence stated in code, config, METRICS.md and the ADR: absolute GED values are not comparable to published Break GED.** Same deviation the survey declared for its own §3 figures.
2. **Per-item budget, and what happens on exhaustion.** Break wraps the optimizer in `@exit_after(180)` and its aggregator **drops** a timed-out item — unacceptable here, because a dropped item has no pair and the whole battery is paired. So: `break_metrics.ged.max_nodes_for_optimizer` (**16**) skips the search above the cap and reports the cost of one concrete edit path (a valid upper bound, computed in one pass, identical on every machine), and `per_item_time_budget_seconds` (20.0) is a **backstop, not a timeout** — the deadline is only tested *between* the optimizer's approximations, so it cannot interrupt one (including the first), which is why the node cap is the guard that actually bounds cost. Either way the item is **reported and flagged** (`ged_fallback`, `ged_fallback_counts`, and `ged_fallback_seconds` for the one machine-dependent path), never dropped. The cap comes from measurement **with the gold hop depth stated**, since the gold's size is part of the cost: against the 4-hop fixture gold, repeated-label predictions cost 0.14 s at 8 nodes, **1.81 s at 16**, 4.05 s at 20 and 18.44 s at 30, while the densest graph measured (30 nodes, 435 edges) took 0.06 s — cost is label ambiguity × nodes × gold size, **not** edges, so an edge bound was considered and rejected. At 17 nodes the deterministic bound equalled the optimizer's value on both repeated-label shapes (+0.0071 on a chain), so what the cap trades away on junk-long predictions is third-decimal. Full table in ADR 0026.

3. **Direction.** Prediction first, gold second (as the official driver calls it); edges run from the referencing step to the referenced one (official `(i+1, ref)`); lower is better, carried in the data as above.

## Known blind spot, pinned rather than hidden

A **2-step** reversal scores `ged` **0.0**: the reversal turns step 1's `#1` into a self-loop, and networkx prices that self-loop against an ordinary edge at 0 (the true minimum edit path costs 2 of 3). A 3-step reversal is priced correctly (0.4 on the fixture). The official evaluator computes GED through the same networkx call, so this is documented and tested rather than "fixed" — fixing it would be inventing a metric. `break_exact_match` catches the reversal in both cases, which is the survey's §4 item 2 in practice.

## Consequence a reviewer should weigh

Existing per-item files under `runs/` (exp-004 … exp-009) lack the new columns, so `--compare` refuses them with the existing "re-run the evaluation to regenerate them" message. Re-scoring those runs is cheap (CPU, from the stored predictions) but it is a **run** with a log entry, and it is not in this PR. `--v1-per-item` (ADR 0020) keeps working: the new columns are required of a v2 file, optional on the v1 path, and what could not be computed is named in `statistics_not_available_in_inputs` plus a `NOT COMPARED` line in the run note.

## Evidence (ran-X-observed-Y, CPU only, in a throwaway worktree)

- `python -m unittest discover -s tests` at the rebased tip → **Ran 449 tests ... OK** (`tests/test_decomposition_metrics.py`: 73, up from main's 49).
- `scripts/smoke_test.py --only decomposition_metric_tests,musique_eval` → **2/2 stages passed**.
- Committed-code run over the committed fixture (4 evaluable rows): existing golden values unchanged (exact match 0.7500, step F1 0.9167, composite 0.9417) and the new columns present — `Break EM: 0.5000 / SARI: 0.9084 / GED: 0.1012 (lower is better) / chain validity: 1.0000`, `fallbacks used: none (… so nothing here is machine-dependent)`, and the new caveat line stating that the GED/SARI **levels** are not comparable to published Break.
- `--compare` over two fixture-derived per-item files (4 aligned items): all 19 rows printed with the `better` column, `ged` row `-0.2095` labelled `lower`, `statistics_not_available_in_inputs: []`, `tests_reported: {bootstrap: 7, mcnemar: 3, paired_t_test: 9}`. Nothing significant, as n = 4 is below the reporting floor — no claim is made from it.
- Gate-1 fix pass (commit 5): **I1** the v2 per-item loader's finite-number gate iterated the *v1* field list, so a `"ged": null` died as a raw `TypeError` and a `"ged": NaN` reached the run-note formatter — the gate now covers the full v2 set (`_NUMERIC_PER_ITEM_FIELDS`, one shared helper for both loaders), with tests for null / NaN / string / bool in a new column plus a set-level test against the same set-difference bug. **I2** the cap timing record measured the wrong axis (2-hop gold, distinct step texts); re-measured, cap 30 → 16, edge bound rejected with the reason. **I3** the run note carries the not-comparable caveat, because the note is what gets quoted into `experiments/log.md`. **Nits:** `time_budget` no longer described as a hard timeout and records its elapsed seconds; `time_budget_no_yield` → `no_optimizer_result` (it never depended on the clock); both knobs validated at load by `_ged_policy` with code-path tests; `name` required in `_paired_t_test_row` so a distance cannot be labelled higher-is-better by omission.
- New hand-computed expected values (arithmetic in each docstring): identical prediction (EM 1, SARI 1, GED 0, chain 1); reversed 3-step (GED 2/5, EM 0, step_f1 1.0); over-long (GED 1/4); empty (GED 1.0, chain 0.0, SARI `((1/9+4/9+5/8+5/7)/4)/3`); no-reference prediction 0.0 on chain validity where the old term gives 1.0; 31-step runaway → GED 32/31 with `ged_fallback: node_cap`.

**No number for any real system is claimed anywhere in this PR** — nothing was run on the ADR 0007 pinned 600, and re-scoring the nine committed arms is a separate logged run.

Rebased onto `a940376`; the only conflict was `docs/adr/README.md`, resolved by keeping main's 0023/0024/0025 rows verbatim and appending the 0026 row (README diff is +1 line).


AhmadiJahid:
Gate 1: full review at e8d7048 (no Critical, 3 Important + 5 Nit) → fix pass f96b3fe → delta-confirm: **CONFIRM, merge unblocked**. Additivity verified stronger than claimed (evaluator run at a940376 vs f96b3fe on the same fixture: zero shared keys changed value). All I1/I2/I3 + nits verified resolved; the numeric gate refuses null/NaN/string/bool naming file/row/field; cap-16 independently supported (0.99 s @16 vs 18.4 s @30 against a 4-hop gold; node_cap items flagged, never dropped, pairing intact).

Residual findings for the next evaluator-touching pass (none blocking):
- Important: the GED cost table is not re-derivable — shapes named, not defined; no committed micro-benchmark, no SHA on the measurements; one headline row (1.81 s @16) did not reproduce under the reviewer's guessed construction. Commit the benchmark that prints the table, or spell each shape's exact step texts and reference pattern in ADR 0026.
- Important: the run-note caveat says "the three numbers above" under a four-number line, and never names the biggest comparability gap for Break EM: these are computed on MuSiQue gold decompositions, not Break's QDMR data. Fix: "four numbers" + one dataset-difference clause.
- Nits: _GED_MIN_NODE_CAP=8 comment should name configs/decomposer_musique.json's stop_after_step_lines as its source; one tautological assertion in the gate-coverage test (spell the four column names literally); the cost table exists in two prose copies (config _note + ADR); lower_is_better_statistics is absent from eval_metrics.json (direction discoverability for machine readers).

PR #45, opened 2026-08-22
Evaluator record fixes: a committed GED cost benchmark, the four-number caveat, and metric direction in eval_metrics.json

AhmadiJahid:
Refs #40

The residual, non-blocking record-quality findings the PR #44 Gate-1 review pinned for "the next evaluator-touching pass". This is that pass. **Additive/record-only: no metric value changes.**

## What changed, item by item

1. **(Important) The GED cost table is re-derivable.** `scripts/ged_cost_benchmark.py` + `configs/ged_cost_benchmark.json` construct every shape explicitly — exact step texts, exact reference pattern — time the evaluator's *own* `_normalized_ged` (node cap lifted) against the committed fixture gold, and print the table plus the commit it was measured at. ADR 0026's table is now that stdout, measured at `b207b33`. `tests/test_ged_cost_benchmark.py` pins each shape's node/edge counts and step texts, and `--max-node-count 8` is the ~2 s smoke path.
2. **(Important) Run-note caveat.** "the three numbers above" → "the four numbers above" (the line above carries four), and the caveat now leads with the comparability gap that matters most for Break EM: these are scored against **MuSiQue gold decompositions, not Break's QDMR annotations**. Pinned by a test.
3. **(Nit)** `_GED_MIN_NODE_CAP`'s comment names `conditions.unguided_capped.stop_after_step_lines` in `configs/decomposer_musique.json` as the source of its 8.
4. **(Nit)** The gate-coverage test spells the four issue #40 columns literally. The old assertion differenced the gate against `_ISSUE_40_STATISTICS`, which is the same constant `_REQUIRED_V1_PER_ITEM_FIELDS` is *derived* by subtracting — emptying the constant would have satisfied it.
5. **(Nit)** `configs/musique_eval.json`'s `break_metrics._note` no longer duplicates the cost table; ADR 0026 is the single canonical copy and the config points at it.
6. **(Nit)** `eval_metrics.json` gains `lower_is_better_statistics` (same source constant as the comparison path, with a `metric_definitions` entry), so a machine reader can discover that `ged_macro` points the other way.

## Evidence

**Additivity — ran the evaluator on the same committed fixture at main (`8603e97`) and at this branch:**

```
scripts/musique_decompositions_evaluator.py --predictions tests/fixtures/predictions/decomposer_results_musique.json
(QAV2_PATHS_CONFIG=configs/smoke_paths.json, separate worktrees, 4 evaluated rows each)
```

- 160 shared leaf keys compared in `eval_metrics.json`; **0 metric values changed**. The only 5 diffs are provenance: `predictions_path` and `gold_path` (the two worktrees' absolute paths), `git.branch`, `git.dirty`, `git.dirty_file_count`.
- Keys only in the branch: `lower_is_better_statistics[0]` and `metric_definitions.lower_is_better_statistics`. None only in main.
- `eval_per_item.json`: 4 rows × 31 columns, **byte-identical values, 0 changed cells**.

**Tests:** `python -m unittest discover -s tests` → **Ran 458 tests … OK** (main is 449; +7 for the new benchmark module, +2 for the new evaluator tests).

**Smoke:** `scripts/smoke_test.py` → 40/40 stages passed, including `decomposition_metric_tests` and `musique_eval` with their golden metric values unchanged.

**Benchmark:** ran at the clean committed SHA `b207b33` (LittleGazor, AMD EPYC 7513, CPU only, networkx 3.6.1, Python 3.13.9); 26 s wall clock for the full table.

## Did the old cost-table numbers reproduce?

Partly, and the ADR says exactly which:

| old row | outcome |
|---|---|
| `gold step texts repeated`, vs 4-hop | **reproduced**: 0.04 / 0.24 / 0.98 / 2.90 / 18.55 s against the old 0.04 / 0.24 / 0.98 / 3.05 / 20.05 s (the two largest cells differ by wall-clock load). This is also the row the Gate-1 reviewer independently reproduced (0.99 s @16, 18.4 s @30). |
| `every step references every earlier step`, vs 4-hop | **reproduced exactly** (0.01 / 0.06 s at 190 / 435 edges); its 30-node cell vs the *2-hop* gold is 0.19 s, above the "≤ 0.06 s" the old table claimed for that column — **corrected**. |
| `repeated step text`, vs 4-hop | **did not reproduce**: ~0.00 s at every size against the old 0.14 / 0.63 / 1.10 / **1.81** / 4.05 / 18.44 s. This is the row the reviewer flagged. Deleted, not kept. |
| `chain-shaped`, vs 4-hop | **did not reproduce**: 0.00 s @12 and 0.01 s @39 against the old 0.07 / 1.27 s. Deleted, not kept. |
| bound vs optimizer at the boundary | **corrected**: the bound is tight, +0.0000 on all four cells (including the 39-step chain at 0.9351), against the old "0.0071 above on the chain shape / 0.9620 vs 0.9597". |

**The cap-16 decision is unaffected and is restated on numbers that exist:** the worst case measured at the cap is ~1.0 s against a 4-hop gold versus ~18.6 s at 30 nodes — an order of magnitude inside the 20 s budget. `max_nodes_for_optimizer` stays 16; no config value moved.

One prose claim was corrected as a consequence: the old ADR said cost is driven by "label ambiguity × node count × gold size". The measured table says something narrower — identical labels with *no* references cost ~0 (networkx yields one approximation and stops), the densest graph (435 edges) is among the fastest, and what is expensive is a prediction whose labels near-tie against the *gold's own* labels while carrying references, rising steeply with gold depth.

## Scope notes

- No `experiments/log.md` entry: the benchmark is a tool, not an experiment — it scores no system and touches no run. No GPU.
- `experiments/` and `runs/` untouched. exp-011's committed `eval_metrics.json` files predate `lower_is_better_statistics`; that is expected and changes no value they carry (recorded in ADR 0026).
- No data files added; the benchmark reads the committed 4-row fixture gold.


AhmadiJahid:
Gate 1 review at `0513488` (base `8603e97`), independent, read-only in throwaway worktrees. **Verdict: CONFIRM, merge unblocked.** No `Critical`. 1 `Important` + 2 `Nit`, all record-quality and none of them touching a metric value, a config value or the cap decision.

## What I ran, and what I observed

**Additivity — reproduced, and slightly stronger than the PR body claims.** Ran `scripts/musique_decompositions_evaluator.py --predictions tests/fixtures/predictions/decomposer_results_musique.json` with `QAV2_PATHS_CONFIG=configs/smoke_paths.json` in two separate worktrees, at `8603e97` and at `0513488`, 4 evaluated rows each. Flattened `eval_metrics.json` to leaf keys: **161 shared keys, 0 metric values changed.** The 4 diffs are all provenance — `created_utc`, `git.commit`, `predictions_path`, `gold_path`. Keys only in the branch: `lower_is_better_statistics[0]`, `metric_definitions.lower_is_better_statistics`. **None only in main.** `eval_per_item.json`: 4 rows × 31 columns = **124 cells, 0 changed**; the only wrapper diffs are `created_utc` and the two worktree paths. Checked no consumer whitelists metrics keys (`plot_pool_sweep.py` reads named keys; `metric_definitions` already establishes non-scalar top-level keys), so the additive list is safe.

**The cost table is re-derivable — reproduced.** Ran `scripts/ged_cost_benchmark.py` at `0513488` on LittleGazor (GPU idle, CPU only), 26 s wall clock. Every node/edge count in ADR 0026's table matches exactly; timings match within load noise:

| cell | ADR (`b207b33`) | my run (`0513488`) |
|---|---|---|
| `gold_step_texts_repeated` 8 / 12 / 16 / 20 / 30 vs 4-hop | 0.04 / 0.24 / **0.98** / 2.90 / 18.55 s | 0.03 / 0.24 / **0.97** / 2.85 / 18.28 s |
| `all_pairs_referencing` 30 (435 edges) vs 2-hop / 4-hop | 0.19 / 0.06 s | 0.18 / 0.06 s |
| bound vs optimizer, all four rows | 1.1450 / 0.7586 / 0.8485 / 0.9351, gap +0.0000 | identical to 4 dp, gap +0.0000 |

Measurement SHA checks out: `b207b33` is `0513488`'s parent and `git diff b207b33 0513488 --name-only` is docs only, so the script and config the ADR names are byte-identical to the ones on the head. Every cell in my run came back with `fallback: None` — no `*`, no `†` markers, so the table is all real optimizer timings.

**The ADR's mechanism claim (b) checks out.** Instrumented `nx.optimize_graph_edit_distance` directly with the evaluator's own `_node_subst_cost`: `repeated_step_text` yields **exactly 1** approximation at 16 and at 30 nodes (`chain_shaped` 39: 1), while `gold_step_texts_repeated` 16 yields 4 (27.0 → 25.0 → 23.0 → 21.0, 0.96 s). So the ~0.00 s cells are cheap for the reason the ADR states.

**The deletions are not a dodge.** I looked for a construction of "repeated step text" that reproduces the deleted 0.14 / 0.63 / 1.10 / **1.81** / 4.05 / 18.44 s row. Nearest plausible reading — N copies of one sentence that *carries* a reference — gives **1.02 s @16 and 2.30 s @20** against the 4-hop gold, i.e. roughly half the old numbers and the same order as the surviving row. Nothing I constructed came near 1.81 s at 16. Deleting rows that no committed construction produces, while quoting the old numbers in prose so the record is not erased, is the right call, and it is the honest direction of travel even though it *lowers* the claimed worst case at the cap (1.8 s → 1.0 s).

**Cap-16 stands, independently.** `max_nodes_for_optimizer` is unchanged at 16 (`configs/musique_eval.json:21`). I probed five shapes beyond the committed set at 16 nodes vs the 4-hop gold — identical-text-with-reference (1.02 s), gold-cycled-with-chain-refs (0.01 s), gold-cycled-with-alternating-suffix (**1.11 s**, my worst), unrelated-text-with-reference (1.02 s), unrelated-text-no-reference (0.00 s). Worst case I could find at the cap is **1.11 s against a 20 s budget**. The decision is robust to my adversarial probing, and "an order of magnitude inside the budget" is now the correct hedge where the previous revision's "two orders of magnitude" was not.

**Tests and smoke.** `python -m unittest discover -s tests` → **Ran 458 tests … OK** on the branch; **449 … OK** at `8603e97`. Delta +9, as claimed. `scripts/smoke_test.py` → `stages_total: 40, stages_passed: 40, stages_failed: []`, with `musique_eval`'s golden values (`exact_match_rate` 3/4, `step_f1_macro` 11/12, `composite_score` 113/120) unchanged.

**Item-by-item on the PR #44 residuals.** All six landed: run note reads `CAVEAT on the four numbers above`, the line above it carries four (`Break EM / SARI / GED / chain validity`), the MuSiQue-vs-QDMR clause is present, and `three numbers above` is gone from the generated note (still present at `8603e97`, count 1); `_GED_MIN_NODE_CAP`'s comment names `conditions.unguided_capped.stop_after_step_lines`, which I confirmed is `8` in `configs/decomposer_musique.json:53`; the gate-coverage test spells the four columns literally *and* pins `_ISSUE_40_STATISTICS` against them, so emptying the constant now fails; the config `_note` no longer duplicates the table; `lower_is_better_statistics` uses the same `LOWER_IS_BETTER_STATISTICS` constant at both emit sites (`scripts/musique_decompositions_evaluator.py:2197` comparison, `:2553` scoring). Standard checklist clean: no retrieval/few-shot/eval-set code touched so no leakage surface, no closed-model API anywhere in the new files, no data added (the benchmark reads the 4-row fabricated fixture committed in #22), no model-size code touched, `experiments/` and `runs/` untouched, `Refs #40` present and #40 correctly still open. Agreed that the benchmark is a tool and not an experiment — it scores no system — and the reasoning is disclosed in both the config `_note` and the ADR rather than assumed.

## Findings

### `Important` — the replacement causal claim in ADR 0026 is not supported by the committed table, and a two-line counter-construction falsifies it

`docs/adr/0026-...md:156-161` now asserts: *"Neither does raw label ambiguity on its own … What is expensive is a prediction whose labels **near-tie against the gold's own labels** while carrying references."* Same claim restated at `docs/METRICS.md:218-219` ("how closely the predicted step labels tie with the gold's") and at `configs/musique_eval.json:19` ("driven by how closely the predicted step labels tie with the GOLD's labels").

Ran, through the benchmark's own `_time_optimizer` and `_graph_of` against the same 4-hop fixture gold: 16 and 30 copies of `"Zzz quibble frobnicate [#1]?"` — labels that share nothing with the gold except the rewritten reference token. Observed **1.02 s @16 and 10.13 s @30**, the same order as `gold_step_texts_repeated` (0.97 / 18.28 s). The identical shape with the reference removed: **0.00 s**. The same referenced shape against the *2-hop* gold: **0.00 s**.

So near-tie against the gold's labels is **not necessary**. What the measurements support is: the prediction's steps tying with *each other*, the graph carrying edges at all, and gold depth. The committed shape set cannot separate those from the stated claim, because the only expensive shape is confounded — `gold_step_texts_repeated` is simultaneously gold-derived *and* edge-carrying, while both cheap tied shapes are either edge-free (`repeated_step_text`) or label-distinct (`chain_shaped`, `all_pairs_referencing`).

This is the same defect class the PR is fixing — a causal story asserted past what the table can carry — and ADR 0026 is thesis-writing material, which is why it is `Important` rather than a `Nit`. It changes no metric, no config value, and not the cap. It also lands in three prose copies, which is the duplication nit 5 was about.

*Fix direction:* either narrow the wording to what the table settles (edges present + labels tied among the prediction's own steps + gold depth; gold-label similarity **not isolated**), or add the discriminating shape to `configs/ged_cost_benchmark.json` — tied labels, edge-carrying, text disjoint from the gold — and let the regenerated table decide. Either way, keep one canonical copy of the causal sentence, as the PR did for the table.

### `Nit` — the `*` / `†` legend did not travel with the table into the ADR

`scripts/ged_cost_benchmark.py:573-576` prints a legend line (`*` = the stop fired, so the time is a lower bound; `†` = networkx yielded no approximation, so the cell is not a timing). ADR 0026 pastes the rows with a hand-written preamble and drops the legend. Nothing is mis-read today — every cell in both the ADR's run and mine is unmarked — but a future re-paste with a marked cell lands in the ADR uninterpretable, and strictly the claim "it is the stdout of `scripts/ged_cost_benchmark.py`" is true of the rows and not of the preamble. (Related and weaker: `tests/test_ged_cost_benchmark.py:268-278` pins the upper-bound ordering on a single 6-node cell.)

### `Nit` — convention capture: the rule this PR establishes has no home outside ADR 0026

This PR establishes a genuinely reusable convention: **a number that justifies a config constant is committed as a script that prints its own table plus the SHA it was measured at, and a tool that scores no system takes no `experiments/log.md` entry.** Right now that rule lives inside ADR 0026 (a record about Break metric conventions) and inside `configs/ged_cost_benchmark.json`'s `_note`. The next person who has to justify a constant will look for the rule, not for the Break-metrics ADR. It is owed its own short `docs/adr/` entry — flagging that the record is owed, not writing it. Nothing already in `docs/adr/` was violated by this diff.

AhmadiJahid:
Fix pass for the Gate-1 review's 1 Important + 2 Nits. Head now `1a19ca5` (3 commits on top of `0513488`). **The evaluator was not touched**: `git diff 0513488 1a19ca5 -- scripts/musique_decompositions_evaluator.py` is empty, so the additivity evidence the review reproduced at `0513488` (161 shared keys, 0 metric values changed; 124 per-item cells, 0 changed) still stands unchanged — no metric value can have moved.

## Important — the causal claim: falsification reproduced, claim withdrawn

Reproduced the reviewer's counter-construction through the benchmark's own `_time_optimizer` before changing any prose: 16 / 30 copies of `"Zzz quibble frobnicate [#1]?"` vs the 4-hop gold → **1.04 s / 10.34 s**; drop the reference → **0.00 s**; same referenced shape vs the 2-hop gold → **0.00 s**. Confirmed.

Rather than only narrowing the wording, four shapes were added to `configs/ged_cost_benchmark.json` so the table can separate the factors: **{nonsense label, gold-derived label} × {no reference, one reference}**, label ties held at maximum in all four, the nonsense pair sharing vocabulary and token count and differing in exactly one token. Regenerated at `21a1203` (LittleGazor, CPU only, 50 s):

| shape | 8 | 16 | 30 | vs 2-hop @30 |
|---|---|---|---|---|
| `nonsense_text_repeated_with_reference` (no word shared with gold) | 0.08 s | **1.01 s** | **10.10 s** | 0.00 s |
| `gold_step_text_repeated_with_reference` (label *is* a gold step) | 0.08 s | **1.01 s** | **10.12 s** | 0.00 s |
| `nonsense_text_repeated_no_reference` | 0.00 s | 0.00 s | 0.00 s | 0.00 s |
| `gold_step_text_repeated_no_reference` | 0.00 s | 0.00 s | 0.00 s | 0.00 s |

The two with-reference rows agree to a hundredth of a second, so **gold-label similarity costs nothing** once ties and references are held fixed; the no-reference halves are free, so references are necessary; `chain_shaped` / `all_pairs_referencing` (edges, distinguishable labels, up to 435 edges) stay ≤ 0.19 s, so ties are necessary; both referenced shapes are 0.00 s against the 2-hop gold, so gold depth is the third factor.

**Final causal wording** (one canonical copy in ADR 0026, with the other two pointing at it):

> cost rises with the prediction's step labels tying with **each other**, × the prediction carrying `#k` references at all, × the gold's depth — all three necessary in the measured set, and none of them the same thing as resembling the gold's text.

The earlier claim is **withdrawn**, not softened, in all three copies (ADR 0026, `docs/METRICS.md:212-224`, `configs/musique_eval.json` `break_metrics._note`), and the benchmark's own shape docstrings no longer carry the falsified explanation either.

Two side effects worth flagging, both corrections *against* the previous revision's prose:
- The new star-shaped shape is the **first non-tight bound** at the cap boundary: `nonsense_text_repeated_with_reference` @17 gives optimizer 0.9622 vs reported bound 1.0210, **gap +0.0588**. "The bound is tight everywhere" is replaced by "tight on four of five shapes, second-decimal on the fifth". The bound-ordering test now checks every shape at three sizes instead of one 6-node cell.
- Worst case at the cap is unchanged in substance: **~1.0 s** (0.97 s `gold_step_texts_repeated`, 1.01 s the new star shape), consistent with the reviewer's own adversarial worst of 1.11 s, which the ADR now cites. `max_nodes_for_optimizer` stays **16**; no config value moved.

## Nits

- **Legend travels with the table.** ADR 0026 now carries the script's exact legend above the rows: per-cell stop 300.0 s, `*` = the stop fired so the time is a lower bound, `†` = networkx yielded no approximation so the cell is not a timing — plus "no cell below carries a mark", which is true of this run.
- **Convention captured:** [ADR 0027 — "A Number That Justifies a Config Constant Is a Committed Script, Not Prose"](https://github.com/AhmadiJahid/Thesis---QAv2/blob/feature/40-evaluator-record-fixes/docs/adr/0027-a-number-that-justifies-a-constant-is-a-committed-script.md), indexed in `docs/adr/README.md`, five clauses (script prints the number; the record carries the SHA + machine context, one canonical copy; what the script prints is what the record shows, legends included; prose says only what the table settles; a tool that scores no system takes no `experiments/log.md` entry, is covered by construction tests, and carries a tiny mode). ADR 0026 points forward to it.

## Verification

- `python -m unittest discover -s tests` → **Ran 460 tests … OK** (458 at `0513488`; +2 for the new shape/refusal tests).
- `scripts/smoke_test.py` → `stages_total: 40, stages_passed: 40, stages_failed: []`.
- Measurement provenance: the table names `21a1203`, and `git diff 21a1203 1a19ca5 --name-only` is **only** `docs/adr/0026-...md`, so the script and config the table was measured with are byte-identical to the ones on the head.
- `experiments/` and `runs/` untouched; no data added (the benchmark reads the committed 4-row fixture gold); no `experiments/log.md` entry, per ADR 0027 clause 5.

Not merging — over to the delta-confirm.

AhmadiJahid:
Delta-confirm on `0513488..1a19ca5` (`94b8ea8` shapes+tests, `21a1203` prose withdrawal + ADR 0027, `1a19ca5` table refresh). Scope: the delta only; my full review at `0513488` stands. **Verdict: CONFIRM, merge unblocked.** No `Critical`, no `Important`. The `Important` I raised at `0513488` is **resolved** — the fix went further than a wording change, which is the right instinct. 2 `Nit`s below, neither blocking.

## The `Important` is resolved, and I reproduced the discriminating table

Re-ran `scripts/ged_cost_benchmark.py` at `1a19ca5` (LittleGazor, CPU only, 50 s wall clock, GPU idle). The 2×2 reproduces, and it settles the question the old shape set confounded:

| shape (vs 4-hop gold) | ADR at `21a1203` | my run at `1a19ca5` |
|---|---|---|
| `nonsense_text_repeated_with_reference` 8 / 16 / 30 | 0.08 / **1.01** / 10.10 s | 0.08 / **1.02** / 10.25 s |
| `gold_step_text_repeated_with_reference` 8 / 16 / 30 | 0.08 / **1.01** / 10.12 s | 0.08 / **1.03** / 10.21 s |
| both `*_no_reference` rows, 8 / 16 / 30 | 0.00 s | 0.00 s |
| both with-reference rows vs the **2-hop** gold | 0.00 s | 0.00 s |

Every node and edge count in the regenerated table matches mine exactly; the pre-existing rows match within load noise (`gold_step_texts_repeated` @16 0.97 vs my 0.98, @30 18.28 vs my 18.34). So: a label sharing no word with either gold and a label that *is* one of the gold's steps cost the same to a hundredth, once ties and references are held fixed. Toggling the reference off drops both to 0.00 s. Gold depth turns the same referenced shape from 0.00 s to ~10 s. The withdrawal is what the table says.

I checked the construction is a real control and not a name: `_NONSENSE_STEP_TEXT` / `_NONSENSE_STEP_TEXT_WITH_REFERENCE` differ in exactly the last token, same token count, and `test_the_discriminating_pair_differs_only_in_the_reference` asserts zero non-reference word overlap with the gold and the 0/16 edge toggle. The shared `@@1@@` token is the one unavoidable overlap, and it does not undermine the conclusion — the *gold-derived* half shares its entire label with a gold step and still costs the same, so the extra overlap buys nothing. `_first_referencing_gold_step` raising rather than falling back is the right call: a silent fallback would have turned that shape into its own control.

**The claim is now the same in all three prose copies, and I diffed the wording.** ADR 0026 §4 ("cost rises with the prediction's step labels tying with *each other*, × the prediction carrying `#k` references at all, × the gold's depth — all three necessary in the measured set, and none of them the same thing as resembling the gold's text"), `docs/METRICS.md:218-223`, and `configs/musique_eval.json:19`'s `_note` all say that, all three explicitly negate edge count *and* gold resemblance, and none overreaches past "in the measured set". Grepping the tree for the old wording, the only surviving occurrence of "near-tie against the gold" is `docs/adr/0026-...md:171`, which is the sentence recording the withdrawal — withdrawn in place with what was falsified stated, not quietly rewritten. Correct handling.

## The disclosed side-effect is stated honestly, and cap-16 stands

`nonsense_text_repeated_with_reference` @17 came back **optimizer 0.9622 / bound 1.0210 / gap +0.0588** in my run too — identical to 4 dp. "Tight everywhere" is gone from exactly where it lived: ADR 0026 now reads "on four of these five shapes the bound is **tight** … on the fifth … **0.0588 above** it … the honest statement is second-decimal on a junk-by-construction item rather than 'tight everywhere'", and the old revision's 0.0071 chain-shape gap is explicitly not carried forward. The invariant that matters is preserved (a bound is never *below* the true distance) and the test that was one 6-node cell now runs every shape × 3 sizes. Volunteering a number that weakens your own earlier claim is the behaviour this review wants to see.

Cap-16 remains justified. Worst at the cap in the regenerated table is 1.01 s (mine: 1.03 s), my five adversarial shapes topped out at 1.11 s, against a 20 s budget and ~18.3 s at 30 nodes. `max_nodes_for_optimizer` is still 16 and `per_item_time_budget_seconds` still 20.0 in `configs/musique_eval.json:21-22`.

## Both nits closed

The `*` / `†` legend now sits immediately above the ADR table, with "No cell below carries a mark" — and no cell in my run carried one either (all `fallback: None`). ADR 0027 is written, indexed in `docs/adr/README.md`, marked *Accepted (implementer convention, pending Jahid)*, and it is the general rule rather than a restatement of the GED case; its "Alternatives considered" records why prose-with-full-specification and a log entry were both rejected, and its "existing constants are not retro-fitted" clause keeps it from silently creating work.

## No-re-run integrity: verified by running, not by reading

`git diff 0513488 1a19ca5 -- scripts/musique_decompositions_evaluator.py` is **0 lines**. The delta is 8 files: the benchmark script/config/tests, three prose surfaces, and ADR 0027 + index. `configs/musique_eval.json`'s change is the `_note` string only — the `ged` block is byte-identical. Nothing imports `scripts/ged_cost_benchmark.py` except its own test, and no smoke stage covers it, so no pipeline path can reach the new shapes.

Because a config file did change, I did not stop at inspection: re-ran the evaluator on the committed fixture at `1a19ca5` and diffed against my `0513488` output. **163 leaf keys both sides, no key added or removed, 0 non-provenance values changed; per-item 4 rows × 31 columns, 0 changed cells.** So my fixture additivity evidence from `0513488` transfers, and the smoke result (40/40 at `0513488`, `scripts/smoke_test.py` unchanged) is not stale. Also confirmed no config file is sha256-stamped anywhere in the evaluator (the hashing is on per-item files and item ids), so the `_note` edit cannot reach a snapshot hash.

Tests: `python -m unittest discover -s tests` → **Ran 460 tests … OK**, against 458 at `0513488` and 449 at base. Delta +2, as claimed.

## Findings

### `Nit` — the comparability consequence of a non-tight bound is one sentence short of where a reader needs it

This delta establishes a genuinely new fact: the reported bound can be **0.0588 above** the optimizer's value on a star shape, where the previous revision believed it tight everywhere. Because GED is lower-is-better, an inflated bound *penalises* the item, so an arm that produces more `node_cap` items carries a small systematic upward (worse) `ged_macro` bias relative to an arm that produces fewer. That is the conservative direction and it is visible per run — but `docs/METRICS.md:175` still describes `ged_fallback_counts` only as "how many items did not get the optimizer's own value", which was a complete description while the bound was believed tight and no longer is. *Fix direction:* one clause where `ged_fallback_counts` is defined, saying the substituted value can sit up to ~0.06 above the optimizer's on measured shapes and therefore that a `ged_macro` comparison between two arms with different `node_cap` counts should read the counts beside the number. No metric changes; this is the reader-facing consequence of a number the delta itself measured.

### `Nit` — ADR 0027's point 5 touches a CLAUDE.md tracking-surface rule and should say so specifically

Points 1-4 are pure implementer convention. Point 5 ("a tool that scores no system is not an experiment … takes no `experiments/log.md` entry, no `experiments/<exp-id>/` trail") draws a boundary around a rule that CLAUDE.md states unconditionally ("No experiment exists unless it is in the log"; every experiment produces a config snapshot, a metrics JSON and a run note). I think the carve-out is right and the reasoning in "Alternatives considered" is sound — a tool that scores no system on no eval set does not fit the log's schema, and diluting the log costs it the one property it must keep. But it is a narrowing of a Gate-2-adjacent rule by an implementer record, and the blanket *pending Jahid* in the status line does not flag which clause is the one that needs his nod. *Fix direction:* a sentence under point 5 naming it as the clause that wants Jahid's confirmation, since he is the one whose log it is. Recording it as pending rather than deciding it silently is already the correct move — this is about making the ask legible.

Nothing already recorded in `docs/adr/` is violated by this delta, and ADR 0027 is the convention-capture entry my `0513488` review said was owed.

PR #48, opened 2026-08-23
Decomposition-quality suite as the reported primary; composite frozen as legacy (Refs #40)

AhmadiJahid:
Implements `docs/analysis/2026-08-23-decomposition-quality-metric-specification.md` (commit `a6ab324`), authorized by ADR 0023 item 1 / ADR 0028 item 3 and re-confirmed by Jahid in session 2026-08-23. Recorded as **ADR 0029** (the number main had reserved for this lane). `Refs #40`.

**This PR specifies and reports; it does not adopt.** Whether any of this becomes the *thesis*-primary metric is issue #6 item 5 — Jahid's with his supervisor.

## What changed

**1. The reported primary is an unblended six-term suite** (`decomposition_report_card`): Break EM, SARI, GED (⇓), `chain_validity`, hop-count EM, and the over/under-decomposition pair kept split and never summed. Every term is a macro mean over every evaluated item and carries its direction, range, provenance and `n_items`. All six were already computed at HEAD, so this is **registration, direction metadata, guards and reporting — not new metric mathematics**. The run note prints the whole card, because the note is what gets quoted into `experiments/log.md`.

**2. `ged_macro` is the single number where one is structurally required** — a *member* of the suite, not a function over it — emitted with its three mandatory caveats every time (lower-is-better; order-light and on a 2-step plan order-*blind*; not comparable to published Break GED without spaCy).

**3. `composite_score` and `_REF_RX` are frozen as legacy, byte-identical, and deliberately not repaired.** Marked legacy in code (`COMPOSITE_SCORE_STATUS`, emitted as `composite_score_status`), in `configs/musique_eval.json`, and in `docs/METRICS.md` §4. The reasoning preserved in ADR 0029 and in a comment on `_REF_RX` itself: **even a corrected regex leaves a micro-pooled rate with a model-dependent denominator inside an aggregate-of-aggregates — the defect *class*, not just its instance** — and freezing keeps every committed number (nine arms, exp-010's 33 sweep cells, both v1 re-analysis notes) quotable. Frozen for reproducibility, not because it is correct.

**4. New per-item columns**: `over_decomposition` / `under_decomposition` (so ADR 0017's asymmetry is significance-testable at all — the two rates were aggregate-only), and `decomp_mean` (the gated contingency, item 6). Registered in `MCNEMAR_STATISTICS` / `BOOTSTRAP_STATISTICS` / `T_TEST_STATISTICS`; both direction columns in `LOWER_IS_BETTER_STATISTICS`.

**5. Guards the note calls for**: `chain_validity_gold_unchained_items` (the free-credit branch, counted — 0 on the pinned 600 and on the fixture); `_assert_suite_terms_are_macro_means`, which **refuses the run** if any headline term drifts from the mean of its per-item column; `n_items` per term; and `reference_validity_micro` / `composite_score` named in the card as excluded, with the reason.

**6. `scripts/decomposition_junk_battery.py` + `configs/decomposition_junk_battery.json`** — the six-system battery (J1 empty, J2 question echo = Break's own `Copy`, J3/J4 fixed steps, J5 gold reversed, J6 gold), scored with the evaluator's own `score_item` so there is no second copy of any metric. A **tool, not an experiment** (ADR 0027 point 5): no `experiments/log.md` entry, no GPU, `runs/run.lock` **not taken**, and a tiny mode so the full path is smoke-testable.

## Where this deviates from the note, and why

- **`over_decomposition` is registered lower-is-better too.** §2.5 item 1 names only `under_decomposition`. Left out, every `over_decomposition` comparison row would be stamped `higher_is_better` and `favours` would name the wrong system on it. §2.1 of the same note prints the pair as "under ⇓, over ⇓ but tolerated", so both are registered and ADR 0017's tolerance is carried by the two rates staying split — never by calling more over-decomposition better. Recorded in ADR 0029's alternatives.
- **P4 `suite_win_rate` is not built.** Out of this brief's scope, and it is cross-arm, has no per-item value and applies HELM's *per-metric-across-scenarios* form *across metrics* — the note's own adaptation. Its definition stands if a sweep ever needs it.
- **§2.1 reporting rule 4** ("a junk-baseline block beside the arms") is implemented where §2.5 item 8 puts it — a committed battery script plus tests — not as a block inside every scoring run.
- The three new per-item columns are **computed on every run but not *required*** of a file being compared. Requiring them would refuse nine committed arms and 33 sweep cells at `--compare` (including in-flight exp-014 work) rather than comparing what they carry; absence is named in `statistics_not_available_in_inputs`.

## No metric value changed — the fixture evidence (PR #44/#45 method)

Evaluator run at base (`cff8a30`) vs branch, then every **shared numeric leaf key** compared:

| artifact | numeric shared keys | changed |
|---|---|---|
| fixture scoring `eval_metrics.json` | 113 | **0** |
| fixture `eval_per_item.json` | 105 | **0** |
| fixture `--compare` `compare_metrics.json` | 136 | **3** — and all three are `tests_reported.{bootstrap: 7→8, mcnemar: 3→5, paired_t_test: 9→12}`, i.e. *counts of tests reported*, not metric values |
| **real 600-item `--compare`** (exp-011 `exp004_unguided` vs `exp004_oracle_guided`) | 154 | **0** |

Non-numeric shared keys that changed are exactly: `lower_is_better_statistics` (1→3 entries), two `metric_definitions` / `comparison_definitions` strings, `statistics_not_available_in_inputs` (0→3 on the real compare, correctly naming the columns those files predate), and `git.*` provenance. `composite_score` and every one of its inputs are untouched, on the fixture and on the real 600. A6 also asserts the frozen golden values and `_REF_RX.pattern` in the test suite.

## `decomp_mean`: gate **PASSED**, so it ships

Battery run from the code of `87e3765`, clean tree, seed 42, on the **ADR 0007 pinned 600** — all 15 rows on the same 600 items, with each arm's per-item file *checked* to cover exactly that set (a junk row over 2411 gold rows and an arm row over 600 would not be a comparison).

| system | break EM ⇑ | SARI ⇑ | GED ⇓ | chain ⇑ | hop EM ⇑ | under ⇓ | over ⇓ | decomp_mean ⇑ |
|---|---|---|---|---|---|---|---|---|
| J1 empty | 0.0000 | 0.2911 | 1.0000 | 0.0000 | 0.0000 | 1.0000 | 0.0000 | 0.0582 |
| J2 question echo (`Copy`) | 0.0000 | 0.4676 | 0.9011 | 0.0000 | 0.0000 | 1.0000 | 0.0000 | 0.1133 |
| J3 one fixed step | 0.0000 | 0.3141 | 0.9546 | 0.0000 | 0.0000 | 1.0000 | 0.0000 | 0.0719 |
| J4 three fixed steps | 0.0000 | 0.3262 | 1.0421 | 0.0000 | 0.3333 | 0.3333 | 0.3333 | **0.1412** |
| J5 gold reversed | 0.0000 | 0.9414 | 0.2875 | 0.2328 | 1.0000 | 0.0000 | 0.0000 | 0.5774 |
| J6 gold | 1.0000 | 1.0000 | 0.0000 | 1.0000 | 1.0000 | 0.0000 | 0.0000 | 1.0000 |
| exp005 unguided / unguided_capped | 0.0517 | 0.5979 | 0.4910 | 0.8605 | 0.5200 | 0.1150 | 0.3650 | **0.5085** |
| exp008 full_train (best arm) | 0.1017 | 0.6823 | 0.3094 | 0.9942 | 0.7100 | 0.2567 | 0.0333 | 0.6357 |

(The full 15-row table is in ADR 0029.) **A1 PASS · A2 PASS · A3 ordering holds, margin +0.1173, exempt from A2 by design · A4 PASS · A5 PASS.** The gate number: best junk `decomp_mean` **0.1412** (J4) vs real-arm worst **0.5085** — margin **+0.3673** — with every one of J1–J4 below every one of the nine arms. Not disqualified, so it ships **as a contingency**, reported beside the suite and never as it.

Two things worth reading past the gate. The table **independently reproduces** every junk figure the analyst measured at `efb8530` from a different code path: GED junk floor 0.9011, SARI empty 0.2911, reversed-gold GED 0.2875, chain-validity junk 0.0000, real-arm GED worst 0.4910, chain worst 0.8605, SARI junk max 0.4676 vs real min 0.5849. (J3/J4 SARI differ from that note's 0.3229/0.3361 because the fixed step texts are this repo's, committed in the battery config; the note's were never recorded.) And **J5 is the argument for the suite in one row**: content-perfect, maximally mis-ordered, it scores GED 0.2875 and SARI 0.9414 — better than every real arm on both — and is caught only by `break_exact_match` (0.0000) and `ordered_step_accuracy`.

## Verification

- `python -m unittest discover -s tests` → **Ran 480 tests, OK** (460 on main; +20, all in `tests/test_decomposition_metrics.py`: `TestDecompositionReportCard` ×10, `TestJunkBattery` ×10 covering A1–A6). Existing tests updated for the new registry rather than relaxed.
- `python scripts/smoke_test.py` → **40/40 stages passed** (unchanged count; the new tests run inside the `decomposition_metric_tests` stage).
- `python scripts/decomposition_junk_battery.py` → the table above, exit 0, all verdicts PASS.
- No closed commercial model is anywhere in the scoring loop; every term is string- or graph-level. No data files added. `runs/run.lock` not taken; `experiments/` and `runs/` untouched.

## For the follow-on re-score run

- A re-score buys **only** the new columns: `over_decomposition`, `under_decomposition`, `decomp_mean`, the report card, `chain_validity_gold_unchained_items` and `composite_score_status`. Every pre-existing value, `composite_score` included, is unchanged — proven above on a real 600-item artifact.
- It is **CPU-only from existing `results.json` files** and needs no GPU and no `run.lock` — the exp-011 precedent. Nine arms plus exp-010's 33 sweep cells is the full coverage set.
- Until an arm is re-scored, `--compare` still works on it and names the three missing columns in `statistics_not_available_in_inputs` — so nothing is blocked on the re-score, including exp-014.
- A comparison now reports 8 bootstrap intervals / 5 McNemar / 12 t-tests (was 7/3/9). Read the counts off `tests_reported`; no multiple-comparison correction is applied, as before.

Refs #40

### DIFF
--- a/configs/decomposition_junk_battery.json
+++ b/configs/decomposition_junk_battery.json
@@ -1,0 +1,35 @@
+{
+  "_note": "Config for scripts/decomposition_junk_battery.py, the junk battery behind the acceptance criteria of the decomposition report card (docs/analysis/2026-08-23-decomposition-quality-metric-specification.md section 3.3, ADR 0029). It is a TOOL, not an experiment: it scores no system of the thesis, takes no experiments/log.md entry, needs no GPU and takes no runs/run.lock. The junk systems are baselines built out of the gold column, in the same sense as the 'Copy' row Break prints in its own results table. The metric's parameters (the GED cost guards, the decomp_mean component list and its GED clamp) are deliberately NOT duplicated here - they are read from evaluator_config, so this battery gates the metric the pipeline actually runs rather than a second copy of it.",
+
+  "seed": 42,
+  "evaluator_config": "musique_eval.json",
+  "gold_key": "musique_dev_gold_clean",
+
+  "_eval_set_note": "The junk baselines are scored on the ADR 0007 pinned evaluation set (200 questions per hop depth), because that is the set every committed arm was scored on and A2/A4/A5 are all junk-versus-arm comparisons - a junk row over the full 2411-row dev gold and an arm row over the pinned 600 would not be a comparison at all (CLAUDE.md, evidence discipline). The gold file supplies the decompositions; these files supply which items. The script additionally CHECKS that each real arm's per-item file covers exactly the scored set and refuses otherwise. --eval-set all-gold scores every gold row instead and is not comparable to any arm.",
+  "eval_set": {
+    "questions_key": "musique_eval_questions_template",
+    "hops": [2, 3, 4]
+  },
+
+  "_fixed_junk_steps_note": "The constant step texts J3 and J4 predict for EVERY item. They are here rather than in the script because they are parameters of the measurement and a run's numbers cannot be reproduced without them, and they are deliberately REFERENCE-FREE: a junk system that emitted '#k' would score nonzero chain_validity and blunt the A2 assertion that measures precisely the opposite. They are also deliberately unrelated to any MuSiQue gold text. J3 takes the first, J4 takes the first three. Changing them changes every junk number below, so change them only with the ADR 0029 table.",
+  "fixed_junk_steps": [
+    "What is the first fact?",
+    "What is the second fact?",
+    "What is the third fact?"
+  ],
+
+  "_real_arms_note": "The committed arms the junk baselines are compared against, by their per-item files under the (gitignored) runs root. These are the nine arms exp-011 re-scored under the Break-faithful columns at commit a99b573 - the only committed artifacts that carry every column this battery needs on the ADR 0007 pinned 600. Nothing here is re-scored: the files are read and aggregated, and decomp_mean plus the two direction columns are DERIVED from the columns already in them by the evaluator's own functions. On a machine without runs/, pass --no-real-arms to run the junk half only (A2/A4/A5 are then reported as not evaluated, because each of them is a statement about junk versus a real arm).",
+  "real_arms": [
+    { "name": "exp004_unguided", "per_item": "exp-011/exp004_unguided/eval_per_item.json" },
+    { "name": "exp004_oracle_guided", "per_item": "exp-011/exp004_oracle_guided/eval_per_item.json" },
+    { "name": "exp004_unguided_capped", "per_item": "exp-011/exp004_unguided_capped/eval_per_item.json" },
+    { "name": "exp005_unguided", "per_item": "exp-011/exp005_unguided/eval_per_item.json" },
+    { "name": "exp005_oracle_guided", "per_item": "exp-011/exp005_oracle_guided/eval_per_item.json" },
+    { "name": "exp005_unguided_capped", "per_item": "exp-011/exp005_unguided_capped/eval_per_item.json" },
+    { "name": "exp008_full_train", "per_item": "exp-011/exp008_full_train/eval_per_item.json" },
+    { "name": "exp009_mixed", "per_item": "exp-011/exp009_mixed/eval_per_item.json" },
+    { "name": "exp009_oracle_hop_matched", "per_item": "exp-011/exp009_oracle_hop_matched/eval_per_item.json" }
+  ],
+
+  "paths_config": "paths.json"
+}
--- a/configs/musique_eval.json
+++ b/configs/musique_eval.json
@@ -7,6 +7,7 @@
   "out_prefix": "eval",
   "limit": null,
 
+  "_composite_score_status_note": "LEGACY. composite_score and the [#k]-only _REF_RX behind its reference_validity_micro term are FROZEN, byte-identical to the form every committed number was produced under (nine arms, exp-010's 33 sweep cells, both v1 re-analysis notes). It is NOT the reported primary - the six-term decomposition_report_card is - and it is deliberately NOT repaired: correcting the regex would move every committed value under an unchanged name while leaving the actual defect in place, because even a corrected reference-validity term is a micro-pooled rate with a MODEL-DEPENDENT denominator inside an aggregate-of-aggregates (issue #40, docs/analysis/2026-08-23-decomposition-quality-metric-specification.md section 5 option 1, ADR 0029). Do not change these weights or that regex to 'fix' the metric; the replacement is the suite. Whether the composite is retired outright is issue #6 item 5, Jahid's with his supervisor.",
   "composite_score_weights": {
     "step_f1_macro": 0.4,
     "ordered_step_accuracy_macro": 0.3,
@@ -20,6 +21,21 @@
     "ged": {
       "max_nodes_for_optimizer": 16,
       "per_item_time_budget_seconds": 20.0
+    }
+  },
+
+  "decomposition_suite": {
+    "_note": "The reported primary is an UNBLENDED six-term suite (decomposition_report_card): Break EM, SARI, GED (lower is better), chain_validity, hop-count EM, and the over/under-decomposition pair kept split and never summed. All six were already computed before this block existed; the suite is registration, direction metadata, guards and reporting, not new metric mathematics, so no committed number moves. The suite's MEMBERSHIP is deliberately not a knob: it is a metric definition and lives in the script (DECOMPOSITION_SUITE_TERMS), next to the code that computes each term. What lives here is the one thing that is a parameter - the contingency blend. See docs/METRICS.md section 2.3 and ADR 0029.",
+    "decomp_mean": {
+      "_note": "The BLENDED CONTINGENCY of the specification's section 2.3 (P3), reported beside the suite and never as the suite. It exists for the case where one blended number is required; the recommendation remains the suite plus ged_macro. Equal weights by construction and there is NO weighted variant: GLUE's unweighted average is the only weighting form the survey found a precedent for, and any unequal weighting would need a Dynascore-style unit conversion that nothing in this repo estimates. ged_clamp is a JUDGMENT CALL with no source: GED is a distance, unbounded above (measured above 1.0 for junk predictions), so an additive blend has to clamp it before flipping direction - the term is 1 - min(ged_clamp, ged). Alternatives considered and not taken: 1/(1+ged), exp(-ged), or leaving GED out of the blend (ADR 0029). The blend SHIPS only because scripts/decomposition_junk_battery.py measured every junk baseline below every real arm on it; an additive blend containing SARI hands junk a nonzero floor, which is the same failure family as issue #40, and a junk baseline outranking a real arm disqualifies the blend rather than prompting a weight tweak. Changing components or ged_clamp means re-running that battery.",
+      "components": [
+        "break_exact_match",
+        "sari",
+        "ged",
+        "chain_validity",
+        "hop_count_exact_match"
+      ],
+      "ged_clamp": 1.0
     }
   },
 
--- a/docs/METRICS.md
+++ b/docs/METRICS.md
@@ -352,9 +352,12 @@ The acceptance criteria are tests in
 - **A1** J6 scores the extremum on every term.
 - **A2** every real arm beats J1–J4 on `ged_macro`, `chain_validity_macro` and
   `break_exact_match_rate`.
-- **A3** SARI is **exempt** from A2 by design and its margin is recorded instead: its floor on
-  this data is high (Break's own published `Copy` row scores SARI 0.431), so it separates junk
-  by a small margin from a high floor.
+- **A3** SARI is **exempt from A2's assertion** by design and its margin is recorded instead:
+  its floor on this data is high (Break's own published `Copy` row scores SARI 0.431), so it
+  separates junk by a small margin from a high floor. The *ordering* — every real arm above
+  every junk baseline on `sari_macro` — is still gated: it carries a `passed` like every other
+  verdict and drives the battery's exit code, so an arm that falls below the junk floor is
+  caught rather than reported.
 - **A4** the **hard gate** on `decomp_mean`: every junk baseline must rank below every real
   arm. Failing it disqualifies the blend — it is not a prompt to re-tune weights, because the
   legacy composite's weights were never the disease.
@@ -535,10 +538,17 @@ Compares two runs **on the same evaluation set**. Parameters live in
 - **McNemar** (`_mcnemar`, `_mcnemar_exact_p`) for the five binary metrics `exact_match`,
   `hop_count_exact_match`, `break_exact_match`, `over_decomposition` and
   `under_decomposition` (the last two added with the §2.3 suite, so the ADR 0017 asymmetry
-  can be significance-tested at all): with b = #(a correct, b wrong) and c = #(a wrong, b correct),
+  can be significance-tested at all): with b = the discordant items whose indicator is 1 for
+  a and 0 for b (`discordant_only_in_a`) and c the reverse (`discordant_only_in_b`),
   the exact two-sided p-value is `min(1, 2 * BinomialCDF(min(b, c); b + c, 0.5))`, computed
   with exact integer arithmetic. With no discordant pairs p = 1.0. `significant` is
-  `p < alpha`.
+  `p < alpha`. **What a 1 means follows the row's direction**, so the same two counts are
+  repeated under a direction-specific name and the name is only ever emitted where it is
+  true: on the three higher-is-better rows a 1 is a correct item, so b = #(a correct, b
+  wrong) and the pair also appears as `correct_only_in_a` / `correct_only_in_b`; on the two
+  lower-is-better rate rows a 1 is the **error occurring**, so b = #(a errs, b does not) and
+  the pair appears as `error_only_in_a` / `error_only_in_b` instead. The run note's McNemar
+  cell prints the neutral `b=` / `c=`, with the reading spelled out beneath the table.
 - **How much evidence there is, recorded next to the verdicts.** Every row carries `n` and
   `underpowered`; each McNemar row also carries `min_attainable_p_value` — the smallest p
   its discordant count m = b + c could have produced, `min(1, 2 · 0.5^m)`, reached when all
@@ -582,11 +592,15 @@ Compares two runs **on the same evaluation set**. Parameters live in
   run note says so too.
 - **A compared metric the inputs do not carry is skipped and named**, not silently omitted:
   `statistics_not_available_in_inputs` in the metrics JSON, plus a `NOT COMPARED` line in the
-  run note. On v2 artifacts it is empty (the §2.1 / §2.2 columns are **required** — a
-  per-item file written before them gets the existing "re-run the evaluation to regenerate
-  them" refusal). It is non-empty only for `--v1-per-item` inputs, which predate those
-  columns; computing them there from the stored steps would be a *re-score of v1 output*
-  rather than a comparison of what v1 measured.
+  run note. Two kinds of input reach it, and the note names both: a `--v1-per-item`
+  prior-work file, which predates the §2.1 / §2.2 columns (those columns are **required** of
+  a v2 file — one written before them gets the existing "re-run the evaluation to regenerate
+  them" refusal); and a **v2 file scored before the §2.3 suite existed**, which is missing
+  `over_decomposition` / `under_decomposition` / `decomp_mean` because those three are
+  computed on every run but deliberately *not* required of a compared file (ADR 0029 — nine
+  committed arms and 33 sweep cells predate them, and this is the common case until they are
+  re-scored). Either way, computing the missing columns here from the stored steps would be
+  a *re-score of that input* rather than a comparison of what it measured.
 - **v1 prior-work inputs** (`--v1-per-item`, `--v1-alignment`; ADR
   [0020](adr/0020-prior-work-re-analysis-convention.md)). `--compare` can read v1's
   bare-list per-item files — the format that predates
--- a/docs/METRICS.md
+++ b/docs/METRICS.md
@@ -8,7 +8,8 @@ a bug.
 
 Every metric here is **string-level**: no model is in the loop (`METRIC_DEFINITIONS`
 `not_a_semantic_metric`). Two decompositions that mean the same thing but word a step
-differently score as a mismatch. That still holds for the metrics added in §2.1 and §2.2:
+differently score as a mismatch. That still holds for the metrics added in §2.1, §2.2 and
+§2.3:
 the graph edit distance is computed by `networkx`, a graph library, and every candidate that
 would have put a *model* in the scoring loop is left out — whether the supervisor's
 "no closed commercial model may judge decomposition quality" extends to open-weight scorers
@@ -25,6 +26,13 @@ overall and per gold hop depth, in the same style as `per_gold_hop_metrics` belo
 are string-level too — no model scores anything on either side. The two halves share one
 reading of "a step" (`src/step_lines.py`), pinned against each other by
 `tests/test_answer_musique.py::TestStepReadingMatchesTheDecompositionEvaluator`.
+
+**What leads the report.** The reported primary is the six-term suite of §2.3
+(`decomposition_report_card`) — six unblended numbers, always printed together. The weighted
+`composite_score` of §4 is **legacy**: frozen byte-identical so committed numbers stay
+quotable, and not repaired. Whether any of this becomes the *thesis*-primary metric is
+[issue #6](https://github.com/AhmadiJahid/Thesis---QAv2/issues/6) item 5 — Jahid's decision
+with his supervisor. This file describes what is measured and reported, not what is adopted.
 
 ## 1. How rows are built
 
@@ -74,9 +82,10 @@ Unless stated otherwise, an aggregate is the **macro average**: the metric is co
 item and the per-item values are averaged over evaluated rows (`_aggregate`). Four keys are
 not macro averages: `reference_validity_micro` pools counts across all rows;
 `predicted_hop_distribution` / `gold_hop_distribution` are item counts per hop count;
-`ged_fallback_counts` is an item count per fallback reason (§2.2); and
+`ged_fallback_counts` is an item count per fallback reason (§2.2);
+`chain_validity_gold_unchained_items` is an item count (§2.3); and
 `composite_score` is computed from the aggregate values, not averaged from per-item ones
-(§4).
+(§4). **Every term of the reported suite (§2.3) is a macro average**, and the run asserts it.
 
 | metrics JSON key | definition (function) |
 |---|---|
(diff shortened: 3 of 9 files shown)


AhmadiJahid:
## Gate 1 review — `0bcc397` vs base `cff8a30`

**Verdict: CONFIRM — no `Critical`, merge unblocked.** 2 `Important`, 3 `Nit`. Everything below was
re-derived in two throwaway detached worktrees (`0bcc397`, `cff8a30`); shared HEAD stayed on
`cff8a30`, `runs/run.lock` was not taken, and `experiments/`, `runs/exp-*` and every tmux session
were untouched (the battery reads `runs/exp-011/*/eval_per_item.json` read-only).

### Ran / observed

| what I ran | observed |
|---|---|
| `unittest discover -s tests` at branch | **Ran 480 tests, OK** |
| same at base | **Ran 460 tests, OK** (so +20, as claimed) |
| `scripts/smoke_test.py` at branch | **40/40 stages passed**, exit 0 |
| fixture scoring, base vs branch, every shared numeric leaf key | 114 shared numeric / **0 changed**; 0 keys removed; 139 new keys, all under `decomposition_report_card` / `decomp_mean_*` / `composite_score_status` / `chain_validity_gold_unchained_items` / `metric_definitions` / `per_gold_hop_metrics` — no dataset text among them |
| fixture `eval_per_item.json`, base vs branch | 105 shared numeric / **0 changed**; 12 new leaves, all `over_decomposition` / `under_decomposition` / `decomp_mean` |
| fixture `--compare` (fixture vs a degraded set I built), base vs branch | 153 shared numeric / **3 changed**, and all three are `tests_reported.{bootstrap 7→8, mcnemar 3→5, paired_t_test 9→12}` |
| **real 600-item `--compare`**, exp-011 `exp004_unguided` vs `exp004_oracle_guided`, base vs branch | 155 shared numeric / **0 changed** (the one numeric diff is `git.dirty_file_count`, my own scratch file) |
| **real 750-item `--compare`** of two pre-existing exp-010 sweep cells (`size2000_balanced…raw` vs `size2000_clustered…typed`) at branch | compares fine; `statistics_not_available_in_inputs = [decomp_mean, over_decomposition, under_decomposition]`; counts stay 7/3/9; base-vs-branch **0 metric values changed** |
| AST-level diff of `_composite_score`, `_reference_validity`, `_chain_validity`, `_sari`, `_normalized_ged`, `_break_exact_match`, `_step_prf`, `_ordered_step_accuracy`, `_rouge_l`, `_exact_decomposition_match`, `_normalize_step`, `_break_steps`, `_decomposition_graph` | **all IDENTICAL**; `_REF_RX = re.compile(r"\[#(\d+)\]")` unchanged at `scripts/musique_decompositions_evaluator.py:102` |
| `scripts/decomposition_junk_battery.py` (my own run, seed 42, pinned set) | exit 0; **A1 PASS · A2 PASS · A3 ordering_holds=True (junk max 0.4676 J2, real min 0.5849 exp004 `unguided_capped`, margin +0.1173) · A4 PASS · A5 PASS**; all 15 rows `n = 600` |
| `git diff --stat 87e3765 0bcc397` | one file, the ADR — so the table's cited code SHA is honest and head-identical |

**The junk table reproduced digit-for-digit**, including the gate: best junk `decomp_mean` **0.1412**
(J4) vs real-arm worst **0.5085** (exp005 `unguided` / `unguided_capped`), margin **+0.3673**, every
J1–J4 below every one of the nine arms. It also independently reproduces the analyst's `efb8530`
figures from a different code path: GED junk floor **0.9011**, SARI empty **0.2911**, reversed-gold
GED **0.2875**, chain junk **0.0000**, real GED worst **0.4910**, real chain worst **0.8605**, SARI
junk max **0.4676** vs real min **0.5849**.

**Same-set discipline holds and is enforced, not asserted:** `_assert_same_eval_set`
(`scripts/decomposition_junk_battery.py:161`) refuses an arm whose normalized-question set is not
exactly the scored set, so all 15 rows really are the same 600 items.

### The three changed `--compare` keys are benign

`tests_reported.*` is read by nothing but the run-note prose (`musique_decompositions_evaluator.py:3010`)
and the tests. Grepped the repo: no analysis script, plotting script or orchestrator consumes it; the
committed `experiments/exp-00*/metrics.json` copies keep their own values. They are counts of rows
emitted, not metric values. Confirmed benign.

### Checklist

- **Evaluation leakage** — clean. J5/J6 are gold-derived *anchors*, labelled as such, never entered
  as an arm and never in `experiments/log.md`; the few-shot retrieval pool is untouched by this diff.
  No question is scored against few-shot examples drawn from itself.
- **Closed-model judging** — clean. No `torch`/`transformers`/commercial-API import in the new script;
  every term is string- or graph-level; the note's own §0 restates the supervisor's rejection and the
  code carries `not_a_semantic_metric`.
- **Reproducibility** — seed 42 in `configs/decomposition_junk_battery.json`, threaded via
  `set_global_seed` and recorded in the JSON output; the SHA and dirty flag are printed and written;
  the blend's components and clamp are in committed config, not in the script. No data file added —
  `fixed_junk_steps` are three invented strings, deliberately reference-free and unrelated to any
  MuSiQue gold text.
- **Model-size** — not applicable to this diff; no model is loaded anywhere in it.
- **Item 2, the freeze** — verified byte-identical at AST level (table above), and the golden
  `composite_score = 113/120` plus `_REF_RX.pattern` are pinned by `TestJunkBattery::test_a6_…`. No
  committed number changes meaning: the 33 sweep cells and the nine exp-011 arms all recompute
  identically at `--compare`, proven on real artifacts above.
- **Item 4, the deviation — it is right.** §2.1 of the specification prints the pair as
  "under ⇓, **over ⇓** but tolerated"; §2.5 item 1 simply under-specifies against its own §2.1. The
  ⇓ on `over` is explicit in the authority document, so registering it lower-is-better follows the
  spec rather than departing from it, and the alternative would make `favours` name the system that
  over-decomposes **more** as the better one. It changes no existing comparison (the per-item column
  did not exist before), it contradicts no published reading (grepped `docs/analysis/` and
  `experiments/log.md` — the only prior mentions are the spec's own), and ADR 0017's tolerance is
  carried where the spec puts it: the two rates stay split and are never summed. Recorded in ADR
  0029's alternatives.
- **Item 6, ADR/index** — the reserved `0029` placeholder row is correctly replaced with a linked
  title and status; numbering is gap-free `0001`–`0030` with `0030` present; ADR 0026 is amended in
  place rather than rewritten; ADR 0029 records the item-4 deviation among its alternatives and says
  "specifies, does not adopt" in three places, deferring primary-metric adoption to issue #6 item 5.
- **Convention capture** — this change **establishes** three conventions and all three are already
  recorded, so no ADR debt is owed: (a) freeze-don't-fix for a defective metric, stamped
  `<metric>_status: "legacy"` in the JSON, the config and `docs/METRICS.md`; (b) a new per-item column
  is *computed always but not required at `--compare`*, with absence named in
  `statistics_not_available_in_inputs`; (c) an acceptance battery is a committed script plus tests,
  not prose. It **violates** nothing recorded — ADR 0026's conventions are preserved and amended, ADR
  0027 point 5 is followed (tool, no log entry, no GPU, no lock, tiny mode present).
- `Refs #40` present. Diff touches no `experiments/`, no `runs/`. The gitignored `runs/smoke/` left in
  the shared checkout is confirmed ignored (`.gitignore:38`) and the shared tree is clean — immaterial.

---

## Findings

### `Important` 1 — two of the five McNemar rows are now labelled backwards in the machine-readable output

`over_decomposition` / `under_decomposition` joined `MCNEMAR_STATISTICS`, but `_mcnemar` still emits
the field names and the definition of a higher-is-better metric. On a real branch run:

```
"over_decomposition": { "system_a_rate": 0.0, "system_b_rate": 0.5,
                        "correct_only_in_a": 0, "correct_only_in_b": 2,
                        "direction": "lower_is_better", ... }
```

`correct_only_in_b: 2` here means *b over-decomposed on 2 items where a did not* — i.e. b was
**wrong**, not correct. Three places carry the inversion:

- `scripts/musique_decompositions_evaluator.py:2583-2584` — `"correct_only_in_a" / "correct_only_in_b"`.
- `scripts/musique_decompositions_evaluator.py:2592` — the inline comment *"Every McNemar metric here
  is higher-is-better"*, which this PR makes false.
- `scripts/musique_decompositions_evaluator.py:377-380` — `COMPARISON_DEFINITIONS["mcnemar"]`, still
  *"with b = #(a correct, b wrong) and c = #(a wrong, b correct)"*.

`direction`, `difference` and `favours` are all **correct** on those rows (`_favours` applies the
registry), and the run note's table uses neutral `b=`/`c=` with a `lower` direction column — so no
verdict is inverted anywhere and no number moves. But CLAUDE.md requires the metric's definition to
be stated where it is reported, and ADR 0029 makes these rows part of the reported primary's
significance surface. Fix direction: direction-aware wording for the two count fields (or a neutral
`discordant_only_in_a/b` with the existing pair kept as aliases) and a definition string that says
what b and c mean when the column is an error indicator; delete the stale comment.

### `Important` 2 — the `NOT COMPARED` note line misdescribes the files it will actually be printed for

`scripts/musique_decompositions_evaluator.py:2987`:

> `A v1 prior-work per-item file predates them; computing them here would be a re-score of v1
> outputs, not a comparison.`

I hit this on a genuine v2 pair — two exp-010 sweep cells — where `v1_format_inputs` is `null` and
neither file is v1. Per the PR's own consequence list, this is the *common* path from now on (33
sweep cells, nine arms, in-flight exp-014), so the line will mostly be read on v2 artifacts it
describes wrongly. The code comment at `:1421-1428` and ADR 0029 both already get this right ("a v2
arm scored before ADR 0029"); only the note prose lags. Fix direction: say "an input scored before
these columns existed (a v1 prior-work file, or a v2 file scored before ADR 0029)".

### `Nit` 1 — A3's ordering is asserted only where no real arm exists

`_verdicts` returns A3 without a `passed` key, and the exit-code gate is
`all(v.get("passed", True) …)` (`scripts/decomposition_junk_battery.py:597`), so on the real 600 the
SARI ordering is **recorded, never enforced** — a future arm falling below the junk SARI floor exits
0. `TestJunkBattery::test_a3_…` does assert it, but against two fixture "arms" on 4 items. The
specification's A3 wants the ordering asserted "so a future arm that falls below the junk floor is
caught rather than reported". This is *not* the exemption papering over a failure — the exemption is
correctly scoped to A2 only, the ordering holds with margin +0.1173, and the A2 terms carry the
load — but the catch happens in the wrong place. Fix direction: let `ordering_holds` drive a
non-zero exit (or an explicit WARN) on the real battery.

### `Nit` 2 — uncorrected multiplicity per comparison grew from 19 tests to 25

7/3/9 → 8/5/12, with no multiple-comparison correction, exactly as before. It is disclosed in the
run note, in `comparison_definitions.multiple_comparisons` and in ADR 0029's consequences, so nothing
is hidden. Flagging only so it is a visible consequence rather than a discovered one: whether ADR
0009's protocol still wants no correction at 25 tests is Jahid's call with his supervisor, not the
implementer's.

### `Nit` 3 — the nine arms' report-card values currently live only in an ADR

ADR 0029's table carries new derived numbers *about thesis systems* (e.g. exp008 `full_train`
`decomp_mean` 0.6357, `ged_macro` 0.3094). ADR 0027 point 5 covers this — no model was run, nothing
was re-scored, and the chain of citation is real (exp-011's log row → its per-item files → the
script + SHA). But when the CPU-only re-score lands, that is where these become log-citable, and any
note quoting the suite for an arm should cite that row rather than the ADR.

AhmadiJahid:
## Review fix pass — `f158bbb` (was `0bcc397`)

Both `Important` findings and `Nit` 1 are fixed in one commit, `f158bbb`. **No metric value
changed.** Built in a throwaway worktree; `runs/run.lock` not taken, `experiments/`, `runs/exp-*`
and every tmux session untouched (the battery and the real `--compare` read
`runs/exp-011/*/eval_per_item.json` read-only, through a gitignored symlink in the worktree).

### `Important` 1 — the two error-rate McNemar rows were labelled backwards

The counts are now named for **what a 1 means on the row**, which the row's `direction` decides:

- `discordant_only_in_a` / `discordant_only_in_b` — neutral, correct on every row, emitted on
  **all five**, and what the run note's `b=` / `c=` cell now prints.
- the direction-specific pair is written **only where its name is true**: `correct_only_in_a/b`
  on the three higher-is-better rows (unchanged, so every committed metrics JSON and
  `docs/analysis/2026-08-23-exp-010-truncation-confound-audit.py` keep their key),
  `error_only_in_a/b` on `over_decomposition` / `under_decomposition`. A consumer that reads
  `correct_only_in_a` off an error row now gets a `KeyError` instead of a number meaning the
  opposite of its name.
- `COMPARISON_DEFINITIONS["mcnemar"]` states both readings; the stale comment claiming "Every
  McNemar metric here is higher-is-better" is deleted; `docs/METRICS.md` §5 updated to match.
- The run note gained one line under the table saying what `b` and `c` are and that what a 1
  means follows the `better` column.

As emitted on the fixture `--compare` at `f158bbb` (`favours`/`direction`/`difference` untouched,
as the review found them):

```
"under_decomposition": { "system_a_rate": 0.0, "system_b_rate": 0.25, "difference": -0.25,
  "discordant_only_in_a": 0, "discordant_only_in_b": 1,
  "error_only_in_a": 0, "error_only_in_b": 1, "discordant_pairs": 1, ...
  "direction": "lower_is_better", "favours": null }
"exact_match": { ... "discordant_only_in_a": 2, "discordant_only_in_b": 0,
  "correct_only_in_a": 2, "correct_only_in_b": 0, ... "direction": "higher_is_better", ... }
```

### `Important` 2 — the `NOT COMPARED` line

Now: *"An input scored before those columns existed does not carry them — either a v1 prior-work
per-item file, or a v2 file scored before ADR 0029 added the suite columns. Computing them here
from the stored steps would be a re-score of that input, not a comparison of what it measured."*
Verified in the real 600-item note, which is exactly the v2 path the review hit.
`docs/METRICS.md` carried the same wrong claim ("non-empty only for `--v1-per-item` inputs") and
is corrected in the same commit.

### `Nit` 1 — A3 is gated now

`A3_sari_exempt` carries `passed` + `failures` like every other verdict, the exit gate is
`all(v["passed"] …)` (no `.get(..., True)` default left), and the exemption stays scoped to A2's
junk-separation **assertion**: the margin is still reported, not asserted as separation. Printed
line is now `A3_sari_exempt: PASS — ordering_holds=True (… margin +0.1173) — EXEMPT from the A2
assertion by design; the ordering is gated`. A new test pushes one real arm's `sari_macro` under
the junk maximum and asserts A3 flips to `passed: false`. ADR 0029 and `docs/METRICS.md` updated.

### Refreshed zero-change evidence at `f158bbb`

Same method as before (evaluator run at each SHA, every **shared numeric leaf key** compared;
`git.*` / paths / timestamps excluded). Fixture `--compare` = perfect vs unevenly-degraded
fixture arms, both scored at the SHA under test; real = exp-011 `exp004_unguided` vs
`exp004_oracle_guided`, 600 items.

**base `cff8a30` → `f158bbb`**

| artifact | shared numeric | changed | numeric keys removed |
|---|---|---|---|
| fixture `eval_metrics.json` | 113 | **0** | 0 |
| fixture `eval_per_item.json` | 105 | **0** | 0 |
| fixture `--compare` | 154 | **3** — `tests_reported.{bootstrap 7→8, mcnemar 3→5, paired_t_test 9→12}` | 0 |
| **real 600-item `--compare`** | 154 | **0** | 0 |

(Counts differ by a key or two from the review's 114/105/153/155 only because of which
provenance keys each harness excludes; the verdict — zero metric values changed, and now also
zero numeric keys *removed* against base — is the same.)

**reviewed head `0bcc397` → `f158bbb`** (the delta of this fix alone)

| artifact | shared numeric | changed | numeric keys removed |
|---|---|---|---|
| fixture `eval_metrics.json` | 156 | **0** | 0 |
| fixture `eval_per_item.json` | 117 | **0** | 0 |
| fixture `--compare` | 193 | **0** | **4** — `mcnemar.{over,under}_decomposition.correct_only_in_{a,b}`, the backwards-named keys, replaced by `discordant_only_in_*` + `error_only_in_*` |
| **real 600-item `--compare`** | 154 | **0** | 0 |

The only non-numeric shared value that changed anywhere is `comparison_definitions.mcnemar` (the
corrected definition string). The four removed keys existed only on this unmerged branch — no
committed artifact carries an over/under McNemar row.

### Other checks

| what I ran | observed |
|---|---|
| `python -m unittest discover -s tests` at `f158bbb` | **Ran 482 tests, OK** (480 at `0bcc397`, 460 at base; +2 = the McNemar naming test and the A3 gating test) |
| `python scripts/smoke_test.py` at `f158bbb` | **40/40 stages passed**, exit 0 |
| `python scripts/decomposition_junk_battery.py` at `f158bbb`, clean tree, seed 42, pinned 600 | **exit 0**; `A1 PASS · A2 PASS · A3 PASS (ordering_holds=True, junk max 0.4676 J2, real min 0.5849 exp004 unguided_capped, margin +0.1173) · A4 PASS · A5 PASS`; all 15 rows n = 600 |
| the battery's 15 rows vs ADR 0029's table | 14 rows matched by label, **all eight values identical**; the 15th (J2) differs only in row label between the two tables, values equal (`0.0000 / 0.4676 / 0.9011 / 0.0000 / 0.0000 / 1.0000 / 0.0000 / 0.1133`) |

**No verdict moved once A3 was gated** — A3 was already `ordering_holds=True`, so it prints PASS
and the exit code is still 0. `_composite_score`, `_REF_RX` and the other frozen functions are
untouched by this commit (`git diff 0bcc397 f158bbb -- scripts/musique_decompositions_evaluator.py`
touches only `COMPARISON_DEFINITIONS["mcnemar"]`, `_mcnemar`'s count fields and two run-note
lines). `Nit` 2 and `Nit` 3 are left as the review filed them: the multiplicity question is
Jahid's with his supervisor, and the report-card values become log-citable with the CPU-only
re-score.

Not merged — over to the delta-confirm.

AhmadiJahid:
## Delta-confirm — `0bcc397..f158bbb`

**Verdict: CONFIRM. No `Critical`, no new findings. Both `Important`s and the A3 `Nit` are
resolved; the other two Nits were explicitly not-in-scope and remain as recorded.** My full review
at `0bcc397` stands unchanged. Read-only in a fresh throwaway worktree; shared HEAD is still
`cff8a30` with a clean tree, `runs/run.lock` not taken, `experiments/` / `runs/exp-*` / tmux
untouched (the battery read `runs/exp-011/*` and `runs/pool_sweep/*` read-only).

### Ran / observed

| what I ran | observed |
|---|---|
| `unittest discover -s tests` at `f158bbb` | **Ran 482 tests, OK** (480 at `0bcc397`, 460 at base — the +2 are exactly `test_the_discordant_counts_are_named_for_what_a_one_means` and `test_a3_ordering_is_gated_by_the_exit_code_like_every_other_verdict`) |
| `scripts/smoke_test.py` at `f158bbb` | **40/40 stages passed**, exit 0, no FAIL stage |
| `scripts/decomposition_junk_battery.py` at `f158bbb` | exit 0; **A1 PASS · A2 PASS · A3 PASS · A4 PASS · A5 PASS**; all 15 rows `n = 600`; every cell of the table **identical to my `0bcc397` run** — gate still 0.1412 (J4) vs 0.5085, margin +0.3673, A3 margin +0.1173. No verdict and no number moved. |
| base `cff8a30` → `f158bbb`, fixture `eval_metrics.json` | 114 shared numeric / **0 metric values changed**, **0 keys removed** |
| base → `f158bbb`, fixture `eval_per_item.json` | 105 / **0 changed**, **0 removed** |
| base → `f158bbb`, fixture `--compare` | 153 / **3 changed** — `tests_reported.{7→8, 3→5, 9→12}` only — **0 removed** |
| base → `f158bbb`, **real 600-item** `--compare` (exp-011 `exp004_unguided` vs `exp004_oracle_guided`) | 155 / **0 changed**, **0 removed** |

(My leaf-flattening counts 114/105/153/155 where the PR reports 113/105/154/154 — a convention
difference in which provenance leaves are counted, not a disagreement. The only numeric diff I saw
beyond `tests_reported` in any run was `git.dirty_file_count`, caused by my own scratch paths file.)

### 1. Important 1 — McNemar counts: **resolved, and the polarity is right in both directions**

Called `_mcnemar` at `f158bbb` on hand-built rows covering all five statistics and checked both the
naming and the arithmetic:

```
exact_match            higher_is_better  naming_ok=True counts_ok=True
hop_count_exact_match  higher_is_better  naming_ok=True counts_ok=True
break_exact_match      higher_is_better  naming_ok=True counts_ok=True
over_decomposition     lower_is_better   naming_ok=True counts_ok=True
under_decomposition    lower_is_better   naming_ok=True counts_ok=True
```

`naming_ok` asserts both presence **and** absence: `correct_only_in_*` appears on exactly the three
higher-is-better rows and nowhere else, `error_only_in_*` on exactly the two lower-is-better rows
and nowhere else. `counts_ok` re-derives `#(a==1, b==0)` and `#(b==1, a==0)` independently and
confirms `discordant_only_in_a/_b` carry them on every row. The KeyError claim holds — the pair is
mutually exclusive, not additive. `difference`, `direction` and `favours` are unchanged and still
correct. The stale "Every McNemar metric here is higher-is-better" comment is gone, and
`COMPARISON_DEFINITIONS["mcnemar"]` now states both readings explicitly.

**Nothing in the repo breaks.** The only non-test consumer is
`docs/analysis/2026-08-23-exp-010-truncation-confound-audit.py:149,154`, which calls
`_mcnemar(..., statistics=("exact_match", "hop_count_exact_match"))` and reads
`correct_only_in_a/b` off `hop_count_exact_match` — a higher-is-better row that keeps the key. I ran
that exact call path against the fixed `_mcnemar`: `correct_only_in_a` present, value correct.

The run note now prints neutral `b=` / `c=` from `discordant_only_in_*` with a new line beneath the
table spelling out what a 1 means per direction. Verified on a real branch run.

### 2. Important 2 — `NOT COMPARED` prose: **resolved on the path I actually hit**

Re-ran the exp-010 sweep pair (`size2000_balanced…raw` vs `size2000_clustered…typed`, n = 750,
`v1_format_inputs: null`) at `f158bbb`. The line now reads:

> An input scored before those columns existed does not carry them — either a v1 prior-work per-item
> file, or a v2 file scored before ADR 0029 added the suite columns.

True of that input. `docs/METRICS.md`'s matching claim ("non-empty only for `--v1-per-item` inputs")
is corrected in the same commit and now names the v2-pre-0029 case as the common one until the
re-score lands — which matches the behaviour. The comparison itself still works and still reports
7/3/9.

### 3. Nit 1 (A3) — **genuinely gated, and the exemption is not widened**

`A3_sari_exempt` now carries `passed` + `failures`, and the exit gate is
`all(v["passed"] for v in verdicts.values())` with the `.get(..., True)` default removed
(`scripts/decomposition_junk_battery.py:620-622`) — so a verdict that forgets `passed` now raises
rather than silently passing. I proved the gate bites rather than trusting the shape: I took my own
real 600-item rows and pushed J2's `sari_macro` to 0.99, changing nothing else.

```
clean     -> A3 passed: True  | exit gate all(passed): True
injected  -> A3 passed: False | A2 passed: True | A4 passed: True | exit gate all(passed): False
             failure: sari_macro: junk J2_question_echo (0.9900) ranks at or above
                      real arm exp004_unguided_capped (0.5849)
```

A2 and A4 stay PASS under the injection and `sari_macro` is still absent from `A2_TERMS` — the
exemption remains scoped to A2's junk-*separation assertion* and has not been widened into a
general escape. `TestJunkBattery::test_a3_…` pins the same behaviour from the other direction
(pushing a real arm below the junk max) and asserts `exempt_from_a2` is untouched.

### 4. The four removed keys — **cleanup, not a compatibility break**

Verified the base-relative claim independently rather than reading it off the PR. Against base
`cff8a30`, **`f158bbb` removes zero keys from any artifact** — fixture metrics, fixture per-item,
fixture `--compare` and the real 600-item `--compare` all report `keys only in OLD: 0` (table
above). The four keys that disappear relative to `0bcc397`
(`mcnemar.{over,under}_decomposition.correct_only_in_{a,b}`) were introduced by this same PR and
never existed at base.

I also checked they never escaped into a committed artifact: enumerating every
`experiments/*/metrics.json` in the repo, the McNemar row names present are exactly
`['break_exact_match', 'exact_match', 'hop_count_exact_match']` — all higher-is-better, all of which
keep `correct_only_in_*`. No committed number, and no consumer of one, loses a key.

### Standing from the full review

- Evaluation leakage, closed-model judging, reproducibility (seed 42 threaded and recorded, SHA
  stamped, config-borne parameters), model-size, data-in-git: all still clean — the delta adds no
  data, no model and no new config value.
- The legacy freeze is untouched by this delta (`_REF_RX` and `_composite_score` are not in the diff).
- **Convention capture:** the delta *establishes* one convention worth having and records it in ADR
  0029's consequences — **a metrics-JSON field name must be true of the row it is on; where polarity
  decides the meaning, emit a neutral name on every row and the polarity-specific name only where it
  holds.** That is now in the ADR and in `docs/METRICS.md` §5, so no record is owed. Nothing recorded
  elsewhere is violated; ADR 0026 and ADR 0027 still bind unchanged.
- **Nit 2 (multiplicity 19 → 25 uncorrected tests)** and **Nit 3 (the nine arms' report-card values
  live only in ADR 0029 until a re-score gets a log row)** are unaddressed by design — both are
  Jahid's calls, not the implementer's, and both remain recorded rather than resolved.

Merge unblocked.

